PR #1, opened 2016-07-05
Feature/9533 move js api into a submodule

samdeane:
Connect to BohemianCoding/Sketch#9533.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2016 
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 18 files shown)


LICENSE.md
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2016
```
bomberstudios: We should probably add a 'who' here : )

Source/Application.js
```diff
@@ -0,0 +1,291 @@
...
+    */
+
+    get build() {
+      return this.metadata['build']
+    }
+
+    /**
+        Returns the full version of Sketch that is running
+
+        @return {string} Version and build number as a string, eg "3.6 (15352)".
```
bomberstudios: We're missing a `*/` here, I think

bomberstudios: And now I understand how the documenting is supposed to work, so disregard the comments I made in https://github.com/BohemianCoding/Sketch/pull/9822

header.js
```diff
@@ -0,0 +1,56 @@
...
+This is a prototype Javascript API for Sketch.
+
+The intention is to make something which is:
+
+- native Javascript
+- an easily understandable subset of the full internals of Sketch
+- fully supported by Bohemian between releases (ie we try not to change it, unlike our internal API which we can and do change whenever we need to)
+- still allows you to drop down to our internal API when absolutely necessary.
+
+Comments and suggestions for this API are welcome - send them to developer@sketchapp.com.
```
bomberstudios: developers@sketchapp.com (we should probably set up a forwarder from one address to the other, since I can't remember which one is the one we're using : )

samdeane: Yeah, `developers` sounds wrong to me, the address is for *developer* support, thus it should be developer@sketchapp.com. So let's definitely set up that alias!

bomberstudios: Filed https://github.com/BohemianCoding/Sketch/issues/9901

PR #2, opened 2016-07-12
Code Review: Add esdoc comments to the existing JS API code (#9532)

samdeane:
Code review for Add esdoc comments to the existing JS API code (#9532):

> As new API code is written, esdoc comments should always be added as part of the process.
> 
> However, we also need to run through the existing API code and add comments to it.
> 

Connect to BohemianCoding/Sketch#9532.

### DIFF
--- a/Source/Layer.js
+++ b/Source/Layer.js
@@ -5,12 +5,22 @@
 // All code (C) 2016 Bohemian Coding.
 // ********************************
 
-// ## Imports
-
 import { WrappedObject } from './WrappedObject.js'
 import { Rectangle } from './Rectangle.js'
 
+/**
+  Represents a Sketch layer.
+ */
+ 
 export class Layer extends WrappedObject {
+
+    /**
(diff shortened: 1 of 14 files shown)


Source/Layer.js
```diff
@@ -74,60 +192,115 @@ export class Layer extends WrappedObject {
       return frame
     }
 
+    /**
+        Returns a newly created shape, which has been added to this layer,
+        and sets it up using the supplied properties.
+
+        @param properties {dictionary} Properties to apply to the shape.
```
bomberstudios: Shouldn't this be `@param {dictionary} properties` instead of `@param properties {dictionary}`?

bomberstudios:
Made a small comment, but looks good to merge

Source/Layer.js
samdeane: Yeah, thanks - I thought that there were probably still a few of those lurking...

PR #5, opened 2016-08-02
Code Review: Improve wrapping in JS API (#10163)

samdeane:
Code review for Improve wrapping in JS API (#10163):

> At the moment we've got a function to wrap Sketch objects in JS objects, but it only works if you explicitly tell it what kind of Sketch object you've got. 
> 
> This is fine when you *know*, eg when asking to make a new layer, you know what kind of layer you want. It's not so good though when doing something like iterating a collection of layers, where all you know is that they are layers, but what you actually need is to make a wrapper for each, and have it be a `Text` object for `MSTextLayer`, and `Image` object for `MSBitmapLayer`, and so on.
> 
> This needs to be improved, so that we can give any of the major layer types to `Application` and get back a wrapper object of the right type. 

Connect to BohemianCoding/Sketch#10163.

### DIFF
--- a/Source/Application.js
+++ b/Source/Application.js
@@ -351,16 +351,17 @@ export class Application extends WrappedObject {
 
                 /** @test {Application#wrapObject} */
                 testWrapObject(tester) {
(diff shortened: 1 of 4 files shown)


pieteromvlee:
:+1:

PR #7, opened 2016-09-27
Code Review: Update Johnnie’s text utilities plugin to use the api (#10574)

samdeane:
Code review for Update Johnnie’s text utilities plugin to use the api (#10574):

> I'm finding that the best way to drive the development of the JS API is to try to make things using it.
> 
> @randomsequence has a text utilities plugin which is a bit broken - seems like it might be a good candidate for rewriting to use the API, to see what emerges from that.

Connect to BohemianCoding/Sketch#10574.

### DIFF
--- a/Source/Group.js
+++ b/Source/Group.js
@@ -52,7 +52,7 @@ export class Group extends Layer {
 
      iterate(block, filter = null) {
        var layers = this._object().layers()
-       this._document.iterateWithRawLayers(layers, filter, block)
+       this._document.iterateWithNativeLayers(layers, filter, block)
      }
 
      /**
--- a/Source/Group.js
+++ b/Source/Group.js
@@ -7,92 +7,111 @@
 
 import { Layer } from './Layer.js'
 import { Rectangle } from './Rectangle.js'
+
+
 /**
-  Represents a group of layers.
- */
+Represents a group of layers.
+*/
 
 export class Group extends Layer {
 
-    /**
-      Make a new group object.
 
-      @param group {MSLayerGroup} The underlying model object from Sketch.
-      @param document The document that the group belongs to.
-    */
+  /**
+  Make a new group object.
 
-    constructor(group, document) {
-      super(group, document)
-    }
+  @param {MSLayerGroup} group  The underlying model object from Sketch.
+  @param {Document} document The document that the group belongs to.
+  */
 
-    /**
-        Is this an group?
+  constructor(group, document) {
+    super(group, document)
+  }
 
-        All Layer objects respond to this method, but only Groups or things that inherit from groups return true.
 
-        @return {bool} true for instances of Group, false for any other layer type.
-    */
+  /**
+  Is this an group?
 
-    is_group() {
-      return true;
-    }
+  All Layer objects respond to this method, but only Groups or things that inherit from groups return true.
 
-    /**
-     Return a list of tests to run for this class.
+  @return {bool} true for instances of Group, false for any other layer type.
+  */
 
-     @return {dictionary} A dictionary containing the tests to run. Each key is the name of a test, each value is a function which takes a Tester instance.
-     */
+  is_group() {
+    return true;
+  }
 
-     /**
-         Perform a function for every sub-layer inside this one.
-         The function will be passed a single argument each time it is
-         invoked - which will be an object representing the sub-layer.
 
-         @param {function(layer: Layer)} block The function to execute for each layer.
-     */
+  /**
+  Return a list of tests to run for this class.
 
-     iterate(block) {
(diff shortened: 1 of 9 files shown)


Source/Layer.js
```diff
@@ -7,317 +7,405 @@
...
-    */
+  All Layer objects respond to this method, but only shape layers (rectangles, ovals, paths etc) return true.
 
-    get isShape() { return false; }
+  @return {bool} true for instances of Group, false for any other layer type.
+  */
 
-    /**
-        Is this an image layer?
+  get isShape() { return false; }
```
pieteromvlee: Is this not something you'd put on a base class and then query our traits?

Source/Layer.js
```diff
@@ -7,317 +7,405 @@
...
+  /**
+  Select the layer.
+  This will clear the previous selection. Use addToSelection() if you wish
+  to preserve the existing selection.
+  */
 
-        @param {dictionary} properties Properties to apply to the shape.
-        @return {Shape} the new shape.
-    */
+  select() {
```
pieteromvlee: `select` vs `deselect` behaviour fels inconsistent with how they treat deselecting others layers?

pieteromvlee: Or maybe its fine. Probably is... Ignore me

Source/Layer.js
samdeane: Could do, but this applies to the javascript classes themselves, not to the underlying traits. I'm deliberately simplifying the view in the API, so I didn't necessarily want to expose the traits.

samdeane: I can always change the implementation later.

Source/Group.js
```diff
@@ -6,49 +6,112 @@
...
+
+  /**
+  Is this an group?
+
+  All Layer objects respond to this method, but only Groups or things that inherit from groups return true.
+
+  @return {bool} true for instances of Group, false for any other layer type.
+  */
+
+  is_group() {
```
bomberstudios: Why is this `is_group` when you're using `isGroup` a few lines further down?

Source/Layer.js
```diff
@@ -7,317 +7,405 @@
...
-    */
+  All Layer objects respond to this method, but only Groups or things that inherit from groups return true.
 
-    get isGroup() { return false; }
+  @return {bool} true for instances of Group, false for any other layer type.
+  */
 
-    /**
-        Is this a text layer?
+  get isGroup() { return false; }
```
bomberstudios: See comment above

Source/Layer.js
```diff
@@ -7,317 +7,405 @@
...
+    if (!style) {
+      style = new Style()
     }
 
-    /**
-      Extract the frame to use for a layer from some properties.
-      If the frame wasn't supplied in the properties, we return a default value instead.
+    var fills = properties.fills
+    if (fills) {
+      delete properties["fills"]
```
bomberstudios: I'm confused by this `properties.fills` vs `properties["fills"]`, but something tells me this is one of those stupid things JavaScript does, so feel free to ignore the comment

Source/Style.js
```diff
@@ -0,0 +1,124 @@
...
+// ********************************
+
+import { WrappedObject } from './WrappedObject.js'
+
+/// A solid fill/border.
+const BCFillTypeColor    = 0
+
+/// A gradient fill/border.
+const BCFillTypeGradient = 1
+
```
bomberstudios: I guess 2 and 3 are Radial and Angular gradients? Are they missing for any particular reason?

Source/Style.js
```diff
@@ -0,0 +1,124 @@
...
+      and other more complex border parameters to be specified.
+
+      @param {array} values A list of colors - each one representing a border to create.
+
+    */
+
+    set borders(value) {
+      var objects = []
+      for (var b in value) {
+        var color = MSColor.colorWithSVGString_(value[b])
```
bomberstudios: Leaving this here to plant a seed in your mind: we need a nice JS API to work with colours : )

Source/Group.js
samdeane: thanks, that's a mistake. I think it's been fixed in the unit test branch!

Source/Style.js
samdeane: They are missing right now because the API hasn't implemented them. I'll add them later.

Source/Style.js
samdeane: 👍

samdeane:
> Maybe traits are not accessible in the JS bridge, but those methods can be written on the base class I think so that you dont have to repeat them for each subclass

`Layer` *is* the base class in this case.

PR #9, opened 2016-10-07
Code Review: Add support for layer exporting to the javascript API (#10671)

samdeane:
Code review for Add support for layer exporting to the javascript API (#10671):

> I think this would be the number one request many people have. In more concrete terms, it'd make Principle happy

Connect to BohemianCoding/Sketch#10671.

### DIFF
--- a/Source/Layer.js
+++ b/Source/Layer.js
@@ -279,6 +279,40 @@ export class Layer extends WrappedObject {
   }
 
 
+  // imageDataForRequest(request) {
+  //   var layer = this.sketchObject
+  //   request.immutableDocument = layer.documentData().immutableModelObject
+  //   request.immutablePage = layer.parentPage().immutableModelObject
+  //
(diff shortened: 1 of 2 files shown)


kgn:
Will this be merged in soon? The first plugin I want to try to write with this new api involves exporting assets for all the iOS and Android sizes.

bomberstudios:
It's pending review for some other changes inside Sketch that are required for this to work. It is scheduled for inclusion on v41, so I guess it'll take a few days for this code to be merged on this repo, and a couple of weeks to be merged into a public beta of Sketch.

Will keep you updated : )

kgn:
Ok, I was wondering why it wasn't working when I merged it in :)

samdeane:
It's currently looking like this might not make it into 41, but if so it'll definitely make 42. Will try to keep you updated...

PR #12, opened 2016-10-18
title should be the second argument to displayDialog_withTitle

kgn:
http://developer.sketchapp.com/examples/code/user-interaction/


bomberstudios:
Nice catch, thanks! : )

PR #17, opened 2016-10-19
Change the output directory for the built SketchAPI.js file

kgn:
I changed the output directory to be within the project repository in `Output/SketchAPI.js` so that everything related to this project is contained within this folder. I also added `Output/` to the `.gitignore` file so the resulting built file won't be committed.

There may be a better name other than `Output` for this folder, the main thing I wanted to accomplish with this PR was for the resulting js file to be within the project folder to keep everything together and not mess with the developer's folder structure outside of the project folder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 node_modules/
 Source/generated/
+build/
+Output/
+.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 node_modules/
 Source/generated/
 build/
-Output/
+**/SketchAPI.js
 .DS_Store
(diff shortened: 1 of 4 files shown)


bomberstudios:
I will accept a patch that makes the output folder settable using a config file or a command line option, but not one that modifies the output path in `gulpfile.js`, since we're using that path in our build process for Sketch : )

kgn:
Makes sense, I'll work on these changes.

kgn:
@bomberstudios alright, lemme know what you think of this implementation

bomberstudios:
Awesome, LGTM!

bomberstudios:
Thanks a lot for your contribution : )

kgn:
My pleasure, can't wait to start contributing to the core now that this setup stuff is out of the way 👍

PR #24, opened 2016-10-22
Getting tests running and adding some new methods

kgn:
Now that I'm up and running thanks to @bomberstudios! I had a chance to read over all of the source files and was able to build, point sketch to my new file, and run all of the tests, after commenting out some(#22, #23).

I also implemented some simple functions that are useful, but also to help me get up to speed with the whole system, so all the functions I added have tests.

I added:
`Selection.first` - return the first selected object
`Selection.last` - return the last selected object
`Rectangle.inset` - inset the bounds of the rectangle
`WrapperObject.toString` - returns the sketch object string so test results print with a useful value
`WrapperObject.compare` - added so `WrapperObjects` can be compared directly in `assertEqual`

I also added instructions to the readme with some basic information about writing tests and running them in Sketch.

I also changed `api_version` > `apiVersion` & `full_version` > `fullVersion`. Everything else in the api seems to be camel case, these were the only snake_case functions.

Because `develop` seems to be the latest code I worked off of that branch and merged in the commits that were applied to `master` so those commits are in this PR too. If you'd like me to break up this PR more, or change anything please let me know! Excited to be contributing some small things, hoping to contribute some bigger things soon :) Thanks again to @bomberstudios and the whole Sketch team, I can't wait to see this come together more!

👍

PR #26, opened 2016-10-23
 Expose Application.tester so plugin developers can write tests for their plugins

kgn:
I started working on my first sketch plugin with this API and wanted to be able to use the same testing capabilities for my plugin.

I implemented this as I worked on my first SketchAPI plugin! I wanted to use TDD to make sure the logic of the plugin was sound: https://github.com/kgn/SketchPlugins/blob/master/OrganizeLayers.sketchplugin/Contents/Sketch/OrganizeLayers.js

``` javascript
var tests = {
    "suites" : {
        "OrganizeLayers": {
            "tests" : {
                testOrganizeLayers1(tester) {
                  // test code...
                  // tester.assertEqual(layer.index, 1);
                }
            }
        }
    }
}

log(sketch.tester.runUnitTests(tests));
```

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2016 
+Copyright (c) 2016 Bohemian Coding. 
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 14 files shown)


kgn:
Oh shoot, I didn't consider this PR would include everything from #24, if that is merged into `develop` then this will reduce down to just the last commit. If not, I can strip out the last commit and submit a PR just for that.

mathieudutour:
I'm actually removing the tests entirely from the API build and found a more independent method to create tests for plugins and enable TDD: http://sketchplugins.com/d/511-christmas-present-skpm-test

PR #27, opened 2016-10-24
Expose layers, and mark nativeLayers as 'private'.

kgn:
As I worked on my own plugins using this new api, I encountered the need to get an array of the currently selected layers. I also think this follows the conventions laid out by others parts of the api where `_` denotes the underlying 'private' Sketch object, like `_object`.


mossyblog:
FYI: Merged these changes into #36 as well.

mathieudutour:
I'm going to close this as the new API is completely different

PR #28, opened 2016-10-24
Updated getStringFromUser so that `initial` can be left out of the call.

kgn:
This makes it easier for plugin developers to leave it blank if it's not needed.
Also changed `msg` to `message` so it looks nicer for people reading the docs or source.


mathieudutour:
I'm going to close this as the new API is completely different

PR #30, opened 2016-10-27
Rectangle updates

kgn:
- Allow a rectangle object to be created with no arguments `Rectangle()` -> `{x: 0, y: 0, width: 0, height: 0}`
- Add `maxX` and `maxY`


mossyblog:
Merged this pull request into #36

PR #34, opened 2016-10-28
Readme / API Changes that I think are missing and get in my way :)

mossyblog:
as you need to point the default writes to the appropriate .plist

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ gulp --output Output/SketchAPI.js
 To have Sketch use the .js file you just built, you can run this:
 
 ```
-defaults write com.bohemiancoding.sketch3 SketchAPILocation "/path/to/SketchAPI.js"
+defaults write ~/Library/Preferences/com.bohemiancoding.sketch3.plist SketchAPILocation "/path/to/SketchAPI.js"
 ```
 
 and Sketch will load the external .js file instead of the bundled version.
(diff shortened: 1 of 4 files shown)


plrthink:
@mossyblog There's an issue with indent :)

mossyblog:
Yeah I realised i had 4 spaces instead of 2, will fix

mossyblog:
All good now

plrthink:
but this line is still be indented more than 2 spaces.

https://github.com/mossyblog/SketchAPI/blob/8c0bb2e4cefe00395e6e6093cc97b7b17e253a27/Source/Layer.js#L89

plrthink:
also the whole newPage function block

mossyblog:
Yeah i pulled a totally rookie move just then.. deep apologies, tabs/spacing etc is now returning back to their factory defaults (I work in a team that get religious about this sort of thing and atom was configured differently)...

plrthink:
It's ok. I think you could keep this pr open. The changes are great for the community.

mossyblog:
Well.. i just realised there was a develop branch (didn't see it) which has most of this covered off... (still finding my feet here).

plrthink:
But it seems like the command to overwrite the bundled SketchAPI is still missing? Or I just missing something. Pls correct me.

mossyblog:
Yes, sorry that part is fine... i'll regroup and push a pull request via develop branch and merge changes...

PR #66, opened 2017-11-14
Write plugins for Sketch in TypeScript

pravdomil:
rewrite of API in ts


bomberstudios:
Hi there,

thanks for the effort, but this is a bit unexpected. We have no issue open for this change, and I think something like this needs to be discussed first.

I'm going to close this PR, but feel free to open an issue to discuss the merits of moving all our code to TypeScript, and we'll work from there.

Sorry!

mathieudutour:
We could support a bit more typescript by just having a `.d.ts`, no need to rewrite everything in typescript.

There are some efforts that started here: https://github.com/skpm/skpm/issues/73#issuecomment-338918930, maybe you could collaborate with him to push it a bit further.

pravdomil:
Well, typescript can cover this types of issues:

PR #68, opened 2017-11-23
Code Review: Set up JS tooling for new JS API (#15731)

mathieudutour:
Code review for BohemianCoding/Sketch#15731 (“Set up JS tooling for new JS API”):

Connect to BohemianCoding/Sketch#15731.


Source/Tester.js
```diff
@@ -17,126 +17,116 @@ import { Document } from './Document.js'
...
      */
 
-    assert(condition, description) {
-        if (!condition) {
-            if (!description) description = ""
-            this._testFailures.push(description)
-        }
+  assert(condition, description) {
+    if (!condition) {
+      throw new Error(description || '')
```
mathieudutour: I also change the `assert` methods so that it throws and end the test instead of continuing the test and potentially crashing sketch when there are multiple asserts in a row (out of bounds index and stuff like this)

Source/DataSupplier.js
```diff
@@ -5,70 +5,78 @@
...
-        dataSupplier = AppController.sharedIntance.dataSupplierManager()
-      }
-      super(dataSupplier)
-      
-      this._application = application
-    }
-  
-    /**
+  constructor(dataSupplier, application) {
+    super(dataSupplier || AppController.sharedInstance.dataSupplierManager())
```
mathieudutour: fixed a bug here as well: `s/sharedIntance/sharedInstance`

Source/Artboard.js
```diff
@@ -5,54 +5,49 @@
...
+  }
+
+  /**
+   * Return a list of tests to run for this class.
+   *
+   * @return {dictionary} A dictionary containing the tests to run. Each key is the name of a test, each value is a function which takes a Tester instance.
+   */
+  static tests() {
+    return {
+      tests: {
```
bomberstudios: Double check that changing `"tests"` into `tests` (no quotes) does not break the existing setup for tests (should be easy to check with an Xcode build). Ditto for other similar changes (it's probably harmless, but since we're running these inside something-that-is-not-a-browser, it never hurts to check)

Source/DataSupplier.js
bomberstudios: Nice!

.vscode/extensions.json
```diff
@@ -0,0 +1,8 @@
+{
+  "recommendations": [
+    "dbaeumer.vscode-eslint",
+    "samverschueren.final-newline",
+    "DmitryDorofeev.empty-indent",
+    "esbenp.prettier-vscode"
+  ]
+}
```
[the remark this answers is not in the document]
mathieudutour: yeah it's intentional. The editor config is not super useful since we run `prettier` as a pre-commit hook.

Here it's just to recommend some extensions to make working on the codebase easier. I really liked how desktop/desktop does it: https://github.com/desktop/desktop/blob/master/docs/contributing/tooling.md

Source/Artboard.js
mathieudutour: all good

bomberstudios:
:shipit:

PR #69, opened 2017-11-27
Code Review: SketchAPI design details (#15747)

### LINKED ISSUE #19: Proposal: Move UI functions to a UserInterface class
In my opinion it's strange to have the UI input functions hang off the Application class. Do you guys think it makes more sense to have them be part of a UserInterface class?

Something like this instead?

``` javascript
var sketch = context.api()
sketch.ui.input("Test", "default");
sketch.ui.select("Test", ["One", "Two"], 1);
sketch.ui.message("Hello mum!");
sketch.ui.alert("Title", "message");
```

`select` is the one that really stands out to me, because "select" has a very understood meaning in Sketch, to select an object. In the `develop` branch this function is `getSelectionFromUser` which if you just read `sketch. getSelectionFromUser` is very confusing, my first thought is this is a function to get the user's current layer selection.

### LINKED ISSUE #50: Add the thinkness of border with new API
I'm modifing my plugin, flowmate, to fix errors and update new APIs. It's really hard to understand the new and old APIs because there is not useful example. Your example is just for few cases and most of plugin still use old API (Please add some example to API documents)

And I have one questions. Do you have a plan to change thinkness of shape using new API? I saw the "Style.js" source and the implement is not hard I guess. That's why I don't understand you don't implement thickness (and gradient)

P.S : Is this way to I modify the core script like Style.js and test directly on Sketch app?

mathieudutour:
Code review for BohemianCoding/Sketch#15747 (“SketchAPI design details”):

Goal: get a more semantic and "JS-like" API (but still maintain backward compatibility for a while)

how:
* no more `context.api()`, everything is moved up to `context.api`
* `new context.api.Group()` creates a new group as expected
* you can use `context.api.fromNative()` to get a boxed object

Connect to BohemianCoding/Sketch#15747.
Fix #50, Fix #19


Source/Settings.js
```diff
@@ -0,0 +1,75 @@
...
+    JSON.stringify(value) || 'undefined',
+    key
+  )
+}
+
+const SUITE_PREFIX = 'plugin.sketch.'
+
+/**
+ * Return the value of a global setting for a given key.
+ * @param context The context dictionary passed to the script when it was invoked.
```
robintindale: Are there cases where `context` will change? Otherwise, having to pass the `context` feels a bit pointless, as we already have it baked into `Application`. Would it violate the API design to have `Settings` owned by `Application` and instantiated with `context`?

mathieudutour: Ah I'm writing a post to explain the philosophy behind those changes, give me a day :)

As a quick answer, you can call those methods with `context.api.Settings.settingForKey` so it's not hoisted in the Application anymore

README.md
```diff
@@ -68,7 +66,7 @@ if (obj1 == obj2) {
 }
 
 // this is better - both wrappers might represent the same object
-if (obj1.sketchObject == obj2.sketchObject) {
+if (obj1.isEqual(obj2.sketchObject)) {
```
samdeane: Isn't that second `sketchObject` superfluous?

samdeane: Isn't that second sketchObject superfluous?

README.md
```diff
@@ -78,25 +76,30 @@ if (obj1.sketchObject == obj2.sketchObject) {
 Here's a very simple example script:
 
 ```javascript
-var sketch = context.api()
+var api = SketchAPI
```
samdeane: If we're going to have a global, I'd be inclined to just call it `sketch`. Whatever we pick is bound to clash with someone somewhere, so we might as well choose the thing that will look cleanest when people use it directly.

I think `sketch.version` looks cleaner than `SketchAPI.version`, and for the purposes of scripts that use it, the API _is_ Sketch.

README.md
```diff
@@ -78,25 +76,30 @@ if (obj1.sketchObject == obj2.sketchObject) {
...
 var selection = document.selectedLayers
 var page = document.selectedPage
 
-var group = page.newGroup({
+var Group = api.Group
+var Shape = api.Shape
+
+var group = new Group({
+  parent: page,
   frame: new sketch.Rectangle(0, 0, 100, 100),
```
samdeane: for the purposes of this example, that should be `new api.Rectangle` I think?

Source/Application.js
```diff
@@ -37,19 +31,16 @@ export class Application extends WrappedObject {
    * @return A new Application object.
    */
   constructor(context) {
-    super(context)
+    log('using `api()` is deprecated. Use `api.X` directly')
```
samdeane: this message needs to be consistent with whatever we end up calling the global

README.md
mathieudutour: ah yes good catch

README.md
mathieudutour: :+1:

README.md
mathieudutour: :+1:

Source/Application.js
mathieudutour: I'm not expecting people to use the function form outside `context.api()` so it will probably always be `api` here

mathieudutour:
Yeah I'm tempted to remove all the backward compatibility, it's a lot of code to maintain.

> Not many people are using the v1 API, so I'd be tempted to call this "API v2.0", and require people to ask for it in the manifest.

Hum that would slow down the adoption I think. I'm more leaning towards providing both (the new as a global and the old as `context.api()`). the performance hit is pretty negligible. Do we have any idea of the number of people using the V1? @bomberstudios

samdeane:
I wouldn't consider breaking v1 of the API. That will destroy what little credibility the API idea has. The whole point is we said we wouldn't break it. Having a clean break though, and saying "that was v1, we didn't get it quite right, you can carry on using it but v2 is better" - that's fine I think.

mathieudutour:
that's what i've done:
- the V1 is accessible on `context.api()`
- the V2 is accessible on `sketch`

samdeane:
Ah, gotcha. That sounds good yeah.

mathieudutour:
I started removing all the old code but it's a pain to do without https://github.com/BohemianCoding/SketchAPI/pull/71/ so I'll wait for that :)

Source/components/Page.js
```diff
@@ -0,0 +1,147 @@
...
+
+  remove() {
+    this._object
+      .documentData()
+      .removePages_detachInstances([this._object], true)
+    return this
+  }
+
+  moveToFront() {
+    // doesn't do anything
```
bomberstudios: I know why this is here, but I wonder if it'd be cleaner to have Page extend a more lightweight class than Group, and use that also for Layer and Artboard… (I'm not sure about that either, just thinking aloud here…)

Source/components/Text.js
```diff
@@ -188,81 +132,99 @@ export class Text extends Layer {
...
-    textLayer.lineSpacingBehaviour = lineSpacingBehaviour
-    const baselineOffset = textLayer.firstBaselineOffset()
-    const rect = this.frame
-    rect.y -= baselineOffset - initialBaselineOffset
-    this.frame = rect
-  }
+  set(value) {
+    const object = this._object
+    object.stringValue = value
+    if (!object.nameIsFixed()) {
```
bomberstudios: Don't we deal with this on Sketch's side? Even if you want to do this manually here, either use `object.updateNameFromStorage()` or check the code there, because the logic we use to set the name is not as simple as `name = value`.

package.json
```diff
@@ -1,25 +1,30 @@
 {
   "name": "sketch-api",
-  "version": "1.1.0",
+  "version": "2.0.0",
   "author": "Sam Deane",
```
bomberstudios: Time to add your name to the list, I guess : )

README.md
```diff
@@ -106,10 +108,13 @@ log(selection.isEmpty)
 group.select()
 rect.addToSelection()
 
-sketch.getStringFromUser('Test', 'default')
-sketch.getSelectionFromUser('Test', ['One', 'Two'], 1)
-sketch.message('Hello mum!')
-sketch.alert('Title', 'message')
+var outputString = sketch.UI.getStringFromUser('Test', 'default')
```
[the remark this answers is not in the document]
mathieudutour: :+1:

Source/Selection.js
```diff
@@ -1,33 +1,28 @@
...
    * @param {Page} page The page that the selection relates to.
    */
   constructor(page) {
-    super(page._object)
-    this._page = page
+    this._object = page._object
   }
 
   /**
    * Return the native Sketch layers in the selection.
```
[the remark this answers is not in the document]
mathieudutour: good point, there are a bunch of comments that I haven't touch and that doesn't reflect the reality anymore. Will update

Source/Selection.js
```diff
@@ -37,7 +32,10 @@ export class Selection extends WrappedObject {
    * @return {number} The number of layers that are selected.
    */
   get length() {
-    return this.nativeLayers.count()
+    return this._object
+      .selectedLayers()
```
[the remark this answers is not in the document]
mathieudutour: hum no, `this.layers` will wrap all the layers and that'll be slow.

Source/api/UI.js
```diff
@@ -0,0 +1,87 @@
...
+ * @param msg The prompt message to show.
+ * @param initial The initial value of the input string.
+ * @return The string that the user input.
+ */
+export function getStringFromUser(msg, initial) {
+  const panel = MSModalInputSheet.alloc().init()
+  const result = panel.runPanelWithNibName_ofType_initialString_label_(
+    'MSModalInputSheet',
+    0,
+    initial || '',
```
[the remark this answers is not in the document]
mathieudutour: you could argue that we expect a string so that would be fine

Source/components/Image.js
```diff
@@ -0,0 +1,52 @@
...
+    return this._object.image
+  },
+  set(image) {
+    let nsImage
+    if (typeof image === 'string') {
+      nsImage = NSImage.alloc().initByReferencingFile(image)
+    } else if (image.class && String(image.class()) === 'NSImage') {
+      nsImage = image
+    } else if (image.class && String(image.class()) === 'NSURL') {
+      nsImage = NSImage.alloc().initWithContentsOfURL_(image)
```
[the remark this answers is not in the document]
mathieudutour: that was the existing code. But it's a good point and we should be more careful.

Maybe throw an error instead of crashing? That's would be another PR tho

Source/components/Page.js
```diff
@@ -0,0 +1,147 @@
...
+   * @return {Selection} A selection object representing the layers that the user has selected.
+   */
+  get selectedLayers() {
+    return new Selection(this)
+  }
+
+  /**
+   * Export this layer (and the ones below it), using the options supplied.
+   *
+   * @param {dictionary} options Options indicating which layers to export, which sizes and formats to use, etc.
```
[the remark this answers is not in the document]
mathieudutour: Yeah I was actually thinking about having `export` as a top-level method instead of a method on layers and pages, like `sketch.export(layer, options)` and `sketch.export([layers])`.

What do you think?

Source/components/Page.js
mathieudutour: yeah I think we will need to at some point. This can go in another PR tho, should be transparent

Source/components/Text.js
mathieudutour: that was the existing code

Source/components/Document.js
```diff
@@ -0,0 +1,108 @@
...
+   * that this object represents.
+   *
+   * If you do want to create a new document, you can do so with Application#newDocument.
+   */
+  constructor(document = {}) {
+    if (!document.sketchObject) {
+      const app = NSDocumentController.sharedDocumentController()
+      app.newDocument(Document)
+      // eslint-disable-next-line no-param-reassign
+      document.sketchObject = app.currentDocument()
```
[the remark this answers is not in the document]
mathieudutour: hum it seems to return `undefined` :/

Source/components/Image.js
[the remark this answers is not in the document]
bomberstudios: The crash comes from Objective-C, I believe, not us. I'll open a new PR

PR #72, opened 2017-12-11
make the key in `supplyData` optional

mathieudutour:


SheffieldKevin:
I think we can leave out the key value altogether in SketchAPI interface. Just provide it to the objc method from within the SketchAPI implementation.

PR #78, opened 2017-12-21
add support for symbols

### LINKED ISSUE #73: No wrappers for MSSymbolInstance and MSSymbolMaster
I'm currently having to hack around the lack of wrappers for these two Sketch objects in my plugin (see https://github.com/nefaurk/select-parent-artboards/issues/6), and I'd prefer not to have to.

Ideally, I could write code such as:

```
var artboardForItem = function(document, item) {
  if (item.isArtboard) {
    return item;
  } else if (item.container != null) {
    return artboardForItem(document, item.container);
  } else {
    return null;
  }
};
```

But instead, I'm having to write code such as:

```
var artboardForItem = function(document, item) {
  if (item.isArtboard) {
    return item;
  } else if (item.sketchObject != null && item.sketchObject.className() == "MSSymbolInstance") {
    var parent = item.sketchObject.parentGroup();
    var container = document.wrapObject(parent);
    return artboardForItem(document, container);
  } else if (item.container != null) {
    return artboardForItem(document, item.container);
  } else {
    return null;
  }
};
```

For more info, see: http://sketchplugins.com/d/209-no-mapped-wrapper-for-mssymbolinstance

mathieudutour:
Connect to BohemianCoding/Sketch#15737

- [ ] ~add symbol from lib~
- [x] test nested symbols
- [x] test image overrides
- [x] docs

fix #73


Source/components/Override.js
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * An MSAvailableOverride. This is not exposed, only used by SymbolInstance
+ */
+export class Override extends WrappedObject {}
+
+Override.type = Types.Override
+Override[DefinedPropertiesKey] = { ...WrappedObject[DefinedPropertiesKey] }
+Factory.registerClass(Override, MSAvailableOverride)
+
+Override.define('path', {
```
opsGavin: Would it be better to call this `layerID`?

It looks like it's only used in the tests, but it feels like it could be a point of confusion. If you had a nested text layer the name could be `symbolID/textLayerID_stringValue`. But this `path` is just going to return `textLayerID`.

Likewise, I'd be tempted to rename `id` as `name` - although I don't think that's as confusing.

Source/components/SymbolInstance.js
```diff
@@ -0,0 +1,114 @@
...
+  },
+  set(id) {
+    // we need to find the symbol master and change the master,
+    // it's not enough to just call `this._object.setSymbolID`
+    const master = this._object
+      .parentPage()
+      .documentData()
+      .symbolWithID(id)
+
+    this.parent = master
```
opsGavin: Not sure I understand this - is this to force a refresh? I think the parent for a symbol instance would always be the page (or artboard, group, etc) that contains the symbol instance so I don't think we'd want to set it to be the master.

You might be better off calling `changeInstanceToSymbol:` for this setter instead.

Source/components/SymbolMaster.js
```diff
@@ -0,0 +1,65 @@
...
+SymbolMaster[DefinedPropertiesKey] = { ...Artboard[DefinedPropertiesKey] }
+Factory.registerClass(SymbolMaster, MSSymbolMaster)
+
+SymbolMaster.define('symbolId', {
+  get() {
+    return String(this._object.symbolID())
+  },
+  // Set the symbolID property. This will tell the parent that something has changed, etc.
+  set(id) {
+    this._object.setSymbolID(id)
```
opsGavin: Would we ever do need to do this? It feels a bit dangerous as it would leave all the symbol instances referencing something which is no longer there.

Source/components/SymbolInstance.js
opsGavin: Just read a bit further down - would we be better off having this a readonly property so that they go through `setMaster`?

Source/components/Override.js
mathieudutour: Oh I want to return `symbolID/textLayerID` here. Haven't looked into nested symbols yet.

For the `id`, it's because I need an `id` property since the SymbolInstance inherits from Layer which defines an Id. Hopefully it's just a unique identifier and people won't need to use it

Source/components/SymbolInstance.js
mathieudutour: It's a typo (haven't tested everything yet), it should be `this.master = master` which would calls the `set(master)` defined underneath.

I don't really like having read-only properties, it's not possible to know that it's read-only in javascript.
So I'm usually creating a method (`instance.getSymbolId()`) so that you know that you can't set it or creating a proper property. In this case, it's possible to have a property so I decided to go with it.

But yes, behind the scene it's just a helper for the master setter.

Source/components/SymbolMaster.js
mathieudutour: yeah, probably not. Like I said, I'm always trying to have proper properties and not read-only ones. Will think a bit more about it.

Source/components/Override.js
opsGavin: > Oh I want to return symbolID/textLayerID here.

There's a private method on `MSOverridePoint` called `layerIDPath` that would give you want you want.

mathieudutour:
I split out the image part in #79 for an easier review

Source/components/Document.js
```diff
@@ -62,7 +62,7 @@ export class Document extends WrappedObject {
    * @return {Layer} A layer object, if one was found.
    */
   getLayerWithID(layerId) {
-    const layer = this._object.documentData().layerWithID_(layerId)
+    const layer = this._object.documentData().layerWithID(layerId)
```
bomberstudios: This block of code is repeated verbatim on lines 83:87… maybe an opportunity to refactor it into a method?

Source/components/__tests__/Override.test.js
```diff
@@ -0,0 +1,78 @@
+/* globals expect, test */
```
bomberstudios: Just wanted to say: yay for tests, because they double as documentation : )

Source/components/Document.js
mathieudutour: There is a comment from Sam explaining why :)

Source/components/SymbolInstance.js
```diff
@@ -0,0 +1,115 @@
...
+  set(id) {
+    // we need to find the symbol master and change the master,
+    // it's not enough to just call `this._object.setSymbolID`
+    const parentPage = this._object.parentPage()
+    if (!parentPage) {
+      throw new Error(
+        'A symbol instance needs to be inserted in a page before setting the symbolId'
+      )
+    }
+    const master = parentPage.documentData().symbolWithID(id)
```
opsGavin: Should we protect this? What would happen if we were given an invalid `id` and `master` came back `nil`? Would the instance be left in a bad state or could we crash trying to wrap / unwrap it?

mathieudutour: Yeah that would probably be bad. There are a lot of places where we should probably add more protection and throw JS errors instead of messing the file.

Mind if I do it all in another PR?

opsGavin: > Mind if I do it all in another PR?

Go for it

Source/components/Document.js
bomberstudios: I missed that, it now makes sense. Thanks! : )

docs/Override.md
```diff
@@ -0,0 +1,20 @@
...
+title: Symbol Override
+order: 211
+section: components
+---
+
+```javascript
+var overrides = symbolInstance.overrides
+```
+
+A [Symbol](https://www.sketchapp.com/docs/symbols/) override. This component is not exposed, it is only returns when accessing the `overrides` of a [Symbol Instance](#symbol-instance).
```
hamishmcneill: @mathieudutour should the comment here read "it is only **_returned_** when accessing" or "it only **_returns_** when accessing"

hamishmcneill: As with the other PRs, the use of "it is only returns when" reads a little odd. Would either "it is only returned when" or "it only returns when" be correct?

docs/SymbolInstance.md
```diff
@@ -0,0 +1,75 @@
...
+
+```javascript
+var SymbolInstance = sketch.SymbolInstance
+```
+
+```javascript
+import { SymbolInstance } from 'sketch-api'
+```
+
+A [Symbol](https://www.sketchapp.com/docs/symbols/) instance. It is an instance of [Layer](#layer) so all the methods defined there are available.
```
hamishmcneill: As with the other PRs, the use of "it is only returns when" reads a little odd. Would either "it is only returned when" or "it only returns when" be correct?

PR #84, opened 2018-01-27
use Symbol.iterator for toArray function

pravdomil:
http://2ality.com/2015/02/es6-iteration.html

### DIFF
--- a/Source/dom/utils.js
+++ b/Source/dom/utils.js
@@ -1,17 +1,19 @@
 function toArray(array) {
-    array[Symbol.iterator] = function* () {
-        let index = 0;
-        return {
-            next() {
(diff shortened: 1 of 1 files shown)


mathieudutour:
Why would we want that? An iteratable is not an array. An array is an iteratable but we really want an array here so that we can `map` and `filter` and such

pravdomil:
you can do `map` and `filter` as `[...arr].map`

mathieudutour:
Sure. Or I can use the `toArray` function. I don't understand the point of this PR, every time I'm using `toArray` it's to `map` it so I would have to `[...toArray(arr)].map` every time which seems counter productive.

PR #86, opened 2018-02-01
fix example for API access

DanielRuf:


mathieudutour:
The readme corresponds to the new API that's going to ship with Sketch 49. So it's actually correct ;)

DanielRuf:
So it is a breaking change and it has changed? Is there some reference for all changes?

mathieudutour:
the old API (`context.api()`) will continue to work. We are providing a new API to start with some good foundations.

All the docs are here: https://github.com/BohemianCoding/SketchAPI/tree/develop/docs. They are not deployed yet since Sketch 49 is not released yet.

mathieudutour:
If you want to have a look at the old API, here it is: https://github.com/BohemianCoding/SketchAPI/tree/release/48.2

DanielRuf:
Ok, thanks for the info.

PR #89, opened 2018-02-08
Code Review: Add support for Libraries in SketchAPI (#16238)

mathieudutour:
Code review for BohemianCoding/Sketch#16238 (“Add support for Libraries in SketchAPI”):

- [x] add a `save` method on `Document`
- [x] add a `open` method on `Document`
- [x] add a `Library` component
- [x] a Library can be created from an open Document
- [x] a Library can be created from a path to a sketch file
- [x] we can get the list of the symbols that can be imported from a Library
- [x] we can import a symbol from a Library
- [x] document all of this
- [x] add unit tests for all of this

Connect to BohemianCoding/Sketch#16238.


bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  There are library changes, but not docs. That's OK as long as you're refactoring existing code
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

Source/dom/components/Document.js
```diff
@@ -106,10 +113,92 @@ export class Document extends WrappedObject {
...
+    }
+
+    return Document.fromNative(document)
+  }
+
+  save(path) {
+    let msdocument = this._object
+    const saveMethod =
+      'writeToURL_ofType_forSaveOperation_originalContentsURL_error'
+    if (!msdocument[saveMethod]) {
```
pieteromvlee: That feels like an ugly hack? There's a big difference between `MSDocument` and `MSDocomentData`. If they're conflated in the API into interchangeable object, thats bad!

Source/dom/components/Document.js
```diff
@@ -106,10 +113,92 @@ export class Document extends WrappedObject {
...
+      }
+    }
+
+    return this
+  }
+
+  close() {
+    let msdocument = this._object
+
+    if (!msdocument.close) {
```
pieteromvlee: Same

Source/dom/components/Library.js
```diff
@@ -0,0 +1,141 @@
...
+
+    const libraryController = AppController.sharedInstance().librariesController()
+    const status = libraryController.addAssetLibraryAtURL(libUrl)
+
+    if (status !== 0) {
+      throw new Error(`Error while adding the library: ${AddStatus[status]}.`)
+    }
+
+    // refresh the UI
+    libraryController.notifyLibraryChange(null)
```
opsGavin: I don't think this should be necessary. This notification should be sent as a consequence of calling `addAssetLibraryAtURL`

Source/dom/components/Library.js
```diff
@@ -0,0 +1,141 @@
...
+
+    const lib = toArray(libraryController.userLibraries()).find(l =>
+      l.locationOnDisk().isEqual(libUrl)
+    )
```
opsGavin: This isn't wrong, but the newly added library will alway be first in the list so you could just do `firstObject` instead.

Source/dom/components/Library.js
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  static getLibraries() {
+    const libraryController = AppController.sharedInstance().librariesController()
+    return toArray(libraryController.libraries()).map(
+      Library.fromNative.bind(Library)
+    )
+  }
+
+  static createLibraryFromDocument(document, path) {
```
opsGavin: I think we'd be better giving this method a different name - e.g. `addUserLibraryFromDocument`. I worry that `create` implies that we are doing something to the document to turn it into a library.

opsGavin: It might also be worth not passing in both `document` and `path`. If we made this `addUserLibraryFromDocumentAtPath` it would avoid the dance we're doing saving `MSDocument` / `MSDocumentData`.

Source/dom/components/ShareableObject.js
```diff
@@ -0,0 +1,73 @@
...
+import { getDocumentData } from '../utils'
+
+const ObjectType = {
+  Symbol: 0,
+  LayerStyle: 1,
+  TextStyle: 2,
+  Unknown: 3,
+}
+
+export class ShareableObject extends WrappedObject {
```
opsGavin: I'm not sure, but this might cause confusion. We have an `MSShareableObject` typedef in Sketch which is basically an `MSModelObject` that can be shared (e.g. a style or a symbol). The `MSShareableObjectReference` which this class appears to be wrapping ties a sharable object to a library.

docs/api/Document.md
```diff
@@ -120,6 +127,18 @@ A method to help find the layers in this document which have the given name.
 
 Return an array of [Layer](#layer).
 
+## Get all the Symbol Masters
+
+```javascript
+var symbols = document.getSymbols()
+```
+
+A method to gett all symbol masters defined in the document.
```
opsGavin: small typo

Source/dom/components/Library.js
```diff
@@ -0,0 +1,141 @@
...
+    }
+    return Document.fromNative(this._object.document())
+  }
+
+  getSymbolReferences() {
+    try {
+      const document = this.getDocument()
+
+      return document
+        .getSymbols()
```
opsGavin: I think we only want to be getting local symbols here. This document could also contain foreign symbols from another library and at the moment these will be returned as well.

opsGavin: Also, depending on what's being done with these references we may want the symbol master to come from a local document rather than the library. For instance; lets say you have a library with a red symbol and you import that symbol into your document. You then change the symbol in the library to be blue, but don't update your document. When you go to insert another instance of that symbol into your document you expect the symbol preview to be red not blue. However, if the reference you are using is pointing to the library's version (as it is here) the preview would be blue.

It might make more sense to make this `symbolReferencesForDocument` or conversely have a `symbolReferencesForLibrary` on the document. It would also be worth taking a look at `MSForeignObjectCollector` as this will do most of this matching for you - I think this uses blocks so, if it helps, we could always add a helper method on the obj-C side.

Source/dom/components/Document.js
mathieudutour: > Yeah, I don’t have much of a choice. When running a plugin I get a MSDocument; when I run the tests, I get a MSDocumentData. I don’t want the dev to bother about the difference

Source/dom/components/Library.js
mathieudutour: > I worry that `create` implies that we are doing something to the document to turn it into a library.

I can see where you are coming from. But at the same time, `addUserLibraryFromDocument` doesn't convey that a Library object will be returned.

A `Library` is a different object than a `Document` in the API so I don't think `createLibraryFromDocument` means that it will mutate the document in any way.

> It might also be worth not passing in both document and path.

I'm not sure I understand. and I want a user to be able to create a library from a document that is not yet saved (hence without a document existing at the path)

Source/dom/components/Library.js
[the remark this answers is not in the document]
mathieudutour: You sure? I couldn't find the code path that call it and the only time `addAssetLibraryAtURL` is called,  there is a `reloadSelectingLibrary` next to it 🤔

Source/dom/components/Library.js
mathieudutour: oh and you can't import foreign symbols? I missed that

Source/dom/components/Library.js
mathieudutour: it is wrong if another library was added at the same path and then deleted? (just trying to understand how this is wrong but will use `firstObject` even tho that seems fragile...)

Source/dom/components/Library.js
mathieudutour: > lets say you have a library with a red symbol and you import that symbol into your document. You then change the symbol in the library to be blue, but don't update your document. When you go to insert another instance of that symbol into your document you expect the symbol preview to be red not blue. However, if the reference you are using is pointing to the library's version (as it is here) the preview would be blue.

That's actually what I would expect. If you get a hold of a library, import one of its symbol, I'd expect the symbol to be the one of th

Source/dom/components/ShareableObject.js
mathieudutour: I'm guessing it's only confusing when working all day with `MSShareableObject` and `MSShareableObjectReference`? ;)

Source/dom/components/Library.js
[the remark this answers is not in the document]
opsGavin: Pretty sure. It's a bit of a convoluted path, but it should happen as a consequence of enabling the library. `reloadSelectingLibrary` is just for refreshing the preferences window, making sure the newly added library is selected.

Having said that, all adding and removing libraries currently goes through `MSAssetLibrariesPreferencePane` and it's quite possible that this won't be updated by a `notifyLibraryChange` either.

Source/dom/components/Library.js
opsGavin: If another library had been added at that path but NOT removed from the library controller then `addAssetLibraryAtURL` would fail. In this case when the new document appears at the same location the library controller would find it and the existing library would attach to it.

I'm not saying what you're doing is wrong BTW. I'm just pointing out that it could be simplified as `addAssetLibraryAtURL` will always add to the start of the list if it returns success. I'm happy for you to keep it this way if you're worried that it's fragile though.

Source/dom/components/Library.js
opsGavin: > I'm not sure I understand. and I want a user to be able to create a library from a document that is not yet saved (hence without a document existing at the path)

As I read the code, at the moment you can pass in a path and `nil` as the second argument or a document and a path. But what if you pass in just the document? At the moment it looks like it would try to save with a `nil` path and then fail. If I've already saved the document before that would confuse me.

I don't think it hurts to separate this. If the method only takes a path then it's clear you need to save your document before you add it.

> at the same time, `addUserLibraryFromDocument` doesn't convey that a Library object will be returned.

Maybe it shouldn't be? You could have another method instead which returns the library at a given URL. You could then update this method so that it doesn't throw if there's already a library at that path (`addAssetLibraryAtURL:` returning `MSAssetLibraryControllerAddStatusDuplicate` - maybe the you could just return `false` in that case.) That way the method would only throw an error on more unusual cases (legacy or invalid documents at the path).

At the moment:

```
createLibraryFromDocument( doc1, path )
... followed by ...
createLibraryFromDocument( path )
... or
createLibraryFromDocument( doc2, path )
... or
createLibraryFromDocument( doc1 )
```
Would throw an error. Whereas:
```
doc1.save(path)
addUserLibraryFromDocumentAtPath(path)
lib = libraryWithPath(path)
...
doc2.save(path)
addUserLibraryFromDocumentAtPath(path)
lib = libraryWithPath(path)
```
Would leave everything in a good state regardless or the order in which they are called.

Source/dom/components/Library.js
opsGavin: Not sure what you're saying here?

Source/dom/components/ShareableObject.js
opsGavin: Maybe - but it might be if people are doing class dumps etc.

What about calling it `SymbolReference`?

`MSShareableObjectReference` is a base class which could either be a `MSSymbolMasterReference` or a `MSSharedLayerReference` or a `MSSharedTextReference` (those last two should probably be renamed `MSSharedLayerStyleReference` and `MSSharedTextStyleReference` 😬 ). `MSShareableObjectReference` is useful for us as a generic way to deal with foreign objects, but for the API maybe it should be clear that this is a "symbol" object.

Source/dom/components/Library.js
mathieudutour: > At the moment it looks like it would try to save with a nil path and then fail

That's fine. It's a required argument so if you don't provide it, it should fail.

> it doesn't throw if there's already a library at that path

Good point, it should throw before that, when trying to save the document if there is already one at the same path.

When I look at the use cases:
1) create a document to use as a library
2) use an existing document that is not yet a library
3) use an existing library

So:
1) `createLibraryFromDocument` is perfect for that. And you want the method to throw when there is already a document where you want to save it (while you probably don't want to throw when saving the document directly)
2) Again pretty good. You want it to throw if it's already a library
3) you just don't use `createLibraryFromDocument` and access them directly

I don't see any other use cases that would require splitting this method into 3.

Source/dom/components/Library.js
mathieudutour: I would expect that when I'm accessing a library and importing one of its symbols, it's importing the library's version and not the existing document's version

Source/dom/components/ShareableObject.js
mathieudutour: I don't want to multiply the classes, it makes it super confusing and worrisome when looking at the documentation. How about `ImportableObject` instead?

Source/dom/components/Library.js
opsGavin: Hmm - that's not how Sketch works at the moment (by design).

mathieudutour: Hum I see. I can understand that. Will change.

Is there a way to programmatically bring the changes from an updated library?

Source/dom/components/ShareableObject.js
opsGavin: `ImportableObject` would work 👍

Source/dom/components/Library.js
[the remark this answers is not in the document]
opsGavin: Hmm - no I'm afraid that conclusion is incorrect.

What `libraryController.notifyLibraryChange()` does is send out an `NSNotification` to tell other parts of the system that the library has changed. By "other parts" I mean the symbol insert menus, the symbol popup in the inspector, etc.

However, `MSAssetLibrariesPreferencePane` doesn't listen for this notification - so calling `libraryController.notifyLibraryChange()` won't have any effect on the preference pane at all.

When we add a library (via the prefs pane) we need to update the selection to reflect this - that's what `reloadSelectingLibrary:` does. We could have instead handled the notification to do this but we'd have had to do a fair bit of work to figure out if it was new. At the time that seemed a bit pointless as the preference pane was the only place we expected a library to be added from.

It sounds as though we now need to add a mechanism to the preference pane to respond to `notifyLibraryChange` - want me to do that?

Source/dom/components/Library.js
opsGavin: You want to use `syncForeignObject:withMaster:fromLibrary:` on `MSAssetLibraryController`.

If you wanted I think you could pass in `nil` for the second parameter and (assuming it's the correct library).

This will do the right thing in terms of importing any missing, nested symbols but if the symbol has already been imported it won't update them - you'd have to call `syncForeignObject:withMaster:fromLibrary:` on them to achieve that.

bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  `Source/dom/components/Document.js` changed, but not its tests. That's OK as long as you're refactoring.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/components/ImageData.js` changed, but not its tests. That's OK as long as you're refactoring.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/components/Layer.js` changed, but not its tests. That's OK as long as you're refactoring.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/components/SymbolMaster.js` changed, but not its tests. That's OK as long as you're refactoring.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/enums.js` changed, but not its tests. That's OK as long as you're refactoring.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/utils.js` changed, but not its tests. That's OK as long as you're refactoring.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/components/ImportableObject.js` changed, but not its tests. That's OK as long as you're refactoring.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/components/Library.js` changed, but not its tests. That's OK as long as you're refactoring.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/test-utils.js` changed, but not its tests. That's OK as long as you're refactoring.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/components/ImageData.js` changed, but not its doc. That's OK as long as you're refactoring.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/components/Layer.js` changed, but not its doc. That's OK as long as you're refactoring.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/settings/Settings.js` changed, but not its doc. That's OK as long as you're refactoring.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/test-utils.js` changed, but not its doc. That's OK as long as you're refactoring.
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

Source/dom/components/ImportableObject.js
```diff
@@ -0,0 +1,88 @@
...
+    return wrapNativeObject(importedSymbol)
+  }
+
+  syncLocalDocumentInstancesWithLibrary(document) {
+    const libraryController = AppController.sharedInstance().librariesController()
+
+    const documentData = getDocumentData(document)
+
+    libraryController.syncForeignObject_withMaster_fromLibrary(
+      this._object.foreignObjectCollectionInDocument(documentData),
```
opsGavin: I don't think this is right. `syncForeignObject:withMaster:fromLibrary:` takes a foreign object as it's first parameter but here we seem to be passing in the array of all foreign objects in the library.

If you really want to make `ImportableObject` sync-able you'd have to find the foreign object in the way that we do in `importShareableObjectReference:intoDocument:`. However, I'd suggest a better approach would be to add a `syncWithRemote` method to the symbol.

```
master = symbolRef.importInDocument(mydoc)
...
master.syncWithRemote()
```

The sync method on the master would be implemented something like (Sorry my JavaScript is poor, but hopefully you get the idea):

```
const libraryController = AppController.sharedInstance().librariesController()
const lib = libraryController.libraryForShareableObject( this._object )
const foreignObject = this._object.foreignObject
if (lib && foreignObject) {
  libraryController.syncForeignObject_withMaster_fromLibrary( foreignObject, nil, lib )
}
```

If we're add this to symbol we should handle the case where the symbol is local (`foreignObject` will be `nil`). Also, we can take advantage of the fact that if the second parameter is `nil` the code will find the remote from the library.

Source/dom/components/Library.js
opsGavin: I have to admit this and the `MSDocumentData` / `MSDocument` `save` hack still doesn't sit comfortably with me.

What if we made it `libraryForDocumentAtPath` and have that method either return the existing library at that path or create and return a new one? That would leave you with 2 methods rather than 3, but I think "save this as a new document" and "use this document as a library" are 2 distinct things.

Source/dom/components/__tests__/ImportableObject.test.js
```diff
@@ -0,0 +1,43 @@
...
+    )
+
+    const document2 = new Document()
+
+    const symbolRefs = lib.getImportableSymbolReferencesForDocument(document2)
+
+    expect(symbolRefs.length).toBe(1)
+    expect(symbolRefs[0].id).toBe(master.symbolId)
+
+    const importedMaster = symbolRefs[0].importInDocument(document2)
```
opsGavin: Just a thought here. If the user did:
```
refs = lib.getImportableSymbolReferencesForDocument(document)
master = refs[0].importInDocument(differentDocument)
```
That could potentially leave things in a bad state (e.g. if the symbol was already in `document` and not in `differentDocument` we not do anything and trying to create an instance would result in a missing symbol.

Would it be worth protecting against this by storing the target document on the `ImportableObject`? So instead we could just do:
```
refs = lib.getImportableSymbolReferencesForDocument(document)
master = refs[0].import()
```

Source/dom/components/ImportableObject.js
```diff
@@ -0,0 +1,88 @@
...
+      this._object.sourceLibrary()
+    )
+  }
+}
+
+ImportableObject.type = Types.ImportableObject
+ImportableObject[DefinedPropertiesKey] = {
+  ...WrappedObject[DefinedPropertiesKey],
+}
+// need to check if we have MSShareableObjectReference because it won't be available on jenkins
```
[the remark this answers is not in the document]
mathieudutour: because it's not in the SketchModel or something. I didn't dig into it, I assumed it was the same reason as for why MSDocument is not available.

Source/dom/components/ImportableObject.js
mathieudutour: > I'd suggest a better approach would be to add a syncWithRemote method to the symbol.

Yes good idea. the use case I had for this was to programmatically sync a library. So my thinking was to start from the library (> ImportableObject > sync). But I guess starting from the document works as well

Source/dom/components/__tests__/ImportableObject.test.js
mathieudutour: Good idea

docs/api/Library.md
```diff
@@ -0,0 +1,85 @@
...
+order: 214
+section: components
+---
+
+```javascript
+var Library = require('sketch/dom').Library
+```
+
+A Sketch Library.
+
```
robintindale: perhaps add a note that the properties are read-only, as I think the expectation we established was that those in this table would be read/write?

Source/dom/components/Library.js
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    // otherwise, let's add it
+    const status = libraryController.addAssetLibraryAtURL(libUrl)
+
+    if (status !== 0) {
+      throw new Error(`Error while adding the library: ${AddStatus[status]}.`)
+    }
+
+    const lib = libraryController.userLibraries().firstObject()
```
robintindale: is it good to rely on `userLibraries` being ordered by most recently added?

mathieudutour: Yep, see previous code review :smile:

robintindale: gotcha 👍

ivanortegaalba:
I have found a problem related with this : #178

Source/dom/components/SymbolMaster.js
```diff
@@ -75,6 +75,20 @@ export class SymbolMaster extends Artboard {
     )
     return true
   }
+
+  unlinkFromLibrary() {
```
ivanortegaalba: When you unlink the component. What does it happen with the children? Do they point to the library? And the overrides?

Thanks in advance! I think that this functionality will be very useful :)

mathieudutour: You unlink a Symbol Master si there is no overrides. The nested symbols still point to the library yes, you can unlink them independently

Source/dom/components/SymbolMaster.js
```diff
@@ -48,6 +48,47 @@ export class SymbolMaster extends Artboard {
...
+    )
+    return true
+  }
+
+  unlinkFromLibrary() {
+    const libraryController = AppController.sharedInstance().librariesController()
+    const lib = libraryController.libraryForShareableObject(this._object)
+    if (!lib) {
+      return false
+    }
```
opsGavin: You probably don't need to check for the library here as you can still unlink even if the library is missing. E.g. If I gave you a document with a foreign symbol in it and you didn't have the library.

Source/dom/components/ImportableObject.js
```diff
@@ -109,11 +111,10 @@ ImportableObject.define('objectType', {
   exportable: true,
   importable: false,
   get() {
-    return ImportableObjectType.Symbol
```
ivanortegaalba: What do you think about having a method to retrieve the SymbolMaster if the importable object is a symbol? I can't find an easy way to get the Symbol from the importable object.

PR #90, opened 2018-02-08
Code Review: Temporary solution for deploying the JS API docs (#16539)

mathieudutour:
Code review for BohemianCoding/Sketch#16539 (“Temporary solution for deploying the JS API docs”):

Connect to BohemianCoding/Sketch#16539.


examples/exporting/.gitignore
```diff
@@ -0,0 +1,37 @@
+# Logs
```
bomberstudios: do we really need multiple copies of this file at this level? wouldn't it be saner to move these rules to the top level .gitignore? Also: do all of these still apply?

mathieudutour: all the examples are based on the default `skpm` templates, I wanted to get them as close as possible to what they would when using `skpm`

PR #104, opened 2018-03-02
:bug: fix examples/text-utilities

nichenqin:
the example not work
- iterate `document.selectedLayers` throw `TypeError`: `document.selectedLayers.forEach is not a function`
- the enums exported from api turns into object, so the compare by `===` always failed
```javascript
log(typeof sketch.Types.Text) // object
```
- wrap `fills` and `borders` into `style` object when create new Shape

### DIFF
--- a/examples/text-utilities/src/text-utilities.js
+++ b/examples/text-utilities/src/text-utilities.js
@@ -53,9 +53,7 @@ function addBaselines(layer, fragments) {
       parent: group,
       frame: rect,
       style: {
-        fills: [
-          { color: '#ff000090', fillType: String(sketch.Style.FillType.color) },
-        ],
+        fills: ['#ff000090'],
         borders: [],
       },
     })
@@ -85,7 +83,7 @@ function addLineFragments(layer, fragments) {
       parent: group,
       frame: localRect,
       style: {
-        fills: [{ color, fillType: String(sketch.Style.FillType.color) }],
+        fills: [color],
         borders: [],
       },
     })
--- a/examples/text-utilities/src/text-utilities.js
+++ b/examples/text-utilities/src/text-utilities.js
@@ -52,8 +52,12 @@ function addBaselines(layer, fragments) {
     const shape = new sketch.Shape({
       parent: group,
       frame: rect,
-      fills: ['#ff000090'],
-      borders: [],
+      style: {
+        fills: [
+          { color: '#ff000090', fillType: String(sketch.Style.FillType.color) },
+        ],
+        borders: [],
+      },
     })
   })
 }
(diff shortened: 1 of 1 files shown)


examples/text-utilities/src/text-utilities.js
```diff
@@ -52,8 +52,12 @@ function addBaselines(layer, fragments) {
     const shape = new sketch.Shape({
       parent: group,
       frame: rect,
-      fills: ['#ff000090'],
-      borders: [],
+      style: {
```
mathieudutour: not sure if this change is needed, if a fill is a string, it will assume that it is a `FillType.color`

examples/text-utilities/src/text-utilities.js
```diff
@@ -95,8 +101,8 @@ export function onAddLineFragments(context) {
   const document = sketch.fromNative(context.document)
 
   // Iterate over each text layer in the selection, calling our addLineFragments function
-  document.selectedLayers.forEach(layer => {
-    if (layer.type === sketch.Types.Text) {
+  document.selectedLayers.layers.forEach(layer => {
```
mathieudutour: good catch! I think I'll add a `forEach`, `map` and `reduce` method to the selection so that it reads better.

examples/text-utilities/src/text-utilities.js
```diff
@@ -95,8 +101,8 @@ export function onAddLineFragments(context) {
...
+    if (layer.type === String(sketch.Types.Text)) {
```
mathieudutour: I'm not sure if we should cast to string or just use `==` instead of `===`

nichenqin: I think you should just note this in document and let developer to chose the method to resolve the problem, because most eslint config use rule `eqeqeq`

examples/text-utilities/src/text-utilities.js
nichenqin: Yes, you're right! But it is still necessary to wrap inside `style` object

examples/text-utilities/src/text-utilities.js
mathieudutour: yes indeed

examples/text-utilities/src/text-utilities.js
mathieudutour: Indeed. But `sketch.Style.FillType.color` is a Number and should stay a number. You can either use `sketch.Style.FillType.color` or `'color'`

nichenqin: Yes, it's a miss. Maybe you should also mention it in document, because the fills type in document is `Object[]`

PR #107, opened 2018-03-03
:bug: set font size

nichenqin:
set font size with `textLayer.sketchObject.font = NSFont. systemFontOfSize_ (size)`not work


mathieudutour:
thanks!

PR #110, opened 2018-03-04
getImageFromUser method in UI api

nichenqin:
add a method to get image files path array from user selection


mathieudutour:
Thanks for the PR but I'd rather have UI that doesn't require hooking into sketch as separate packages. For example, you could use https://github.com/skpm/dialog.

PR #112, opened 2018-03-05
Code Review: Fix Artboard.adjustToFit in SketchAPI (#17202)

### LINKED ISSUE #111: adjustToFit method not work for Artboard
It looks like the methods inherits from `Group`, but it just not works.
I find it works by following script
```javascipt
document.sketchObject.actionsController().actionForID("MSResizeArtboardToFitAction").resizeArtboardToFit(nil)
```
but should ensure the `artboard` layer is selected..

Did I miss something or is this a bug? Thanks!

mathieudutour:
Connect BohemianCoding/Sketch#17202

Fix #111

### DIFF
--- a/Source/dom/components/Artboard.js
+++ b/Source/dom/components/Artboard.js
@@ -24,6 +24,15 @@ export class Artboard extends Group {
     }
     super(artboard)
   }
+
+  /**
+   * Adjust the Artboard to fit its children.
+   * override the group's method
+   */
+  adjustToFit() {
+    this._object.resizeToFitChildren()
+    return this
+  }
 }
 
 Artboard.type = Types.Artboard


Source/dom/components/Artboard.js
```diff
@@ -24,6 +24,15 @@ export class Artboard extends Group {
     }
     super(artboard)
   }
+
+  /**
+   * Adjust the Artboard to fit its children.
+   * override the group's method
+   */
+  adjustToFit() {
+    this._object.resizeToFitChildren()
```
nichenqin: is it `resizeToFitChildrenWithOption`?

mathieudutour: no, `resizeToFitChildrenWithOption` doesn't do anything as you notice. This is a new method that I created

nichenqin: this method will not be available until v49.1?

mathieudutour: no I just created it :)

PR #114, opened 2018-03-06
Fix typo in DataSupplier

bomberstudios:
Connect to BohemianCoding/Sketch#17228


mathieudutour:
it's annoying that we can't write tests for it...

PR #115, opened 2018-03-06
Code Review: Dialogs opened by plugins should not show the Sketch icon (#17039)

mathieudutour:
Resolves BohemianCoding/Sketch#17039


Source/ui/UI.js
```diff
@@ -33,8 +33,13 @@ export function message(text, document) {
  * by pressing the OK button.
  */
 export function alert(title, text) {
-  const app = NSApplication.sharedApplication()
-  app.displayDialog_withTitle(text, title)
+  const dialog = NSAlert.alloc().init()
+  dialog.setMessageText(title)
+  dialog.setInformativeText(text)
+  if (__command.pluginBundle() && __command.pluginBundle().icon()) {
+    dialog.icon = __command.pluginBundle().icon()
```
opsGavin: If the plugin hasn't supplied it's own icon would this still default to the sketch icon? Wondering if we need an `else` to set the plugin icon here instead?

Source/ui/UI.js
```diff
@@ -33,8 +33,13 @@ export function message(text, document) {
  * by pressing the OK button.
  */
 export function alert(title, text) {
-  const app = NSApplication.sharedApplication()
-  app.displayDialog_withTitle(text, title)
+  const dialog = NSAlert.alloc().init()
```
opsGavin: BTW - in Obj-C you can do `[class new]` as a shortcut to `[[class alloc] init]`. It doesn't matter, but I expect you could do `NSAlert.new()` here.

Source/ui/UI.js
mathieudutour: Ah yes that'd be great indeed, good call

Source/ui/UI.js
mathieudutour: indeed but I find it confusing with the `new` keyword in javascript :)

opsGavin: Fair enough :)

PR #120, opened 2018-03-08
Bug/set override value

nichenqin:
there is two issues when set override value:

1. If I set a non string value to override whose `property` is `stringValue`(a `number` type value for example), Sketch just crash before we can get any error message.

1. If someone set the value by `setOverrideValue` which is implemented in `SymbolInstance`, the checks defined in `Override` will not work


mathieudutour:
Good catch, thanks!

PR #125, opened 2018-03-10
:bug: addPage is not a function of MSDocument instance

nichenqin:
It throws error when create a page.

And I wonder if it is neccesary to init a page with a frame
`initWIthFram(new Rectange(0, 0, 100, 100).asRect())`


mathieudutour:
Ah indeed. The thing is, `Document` is a wrapper for both `MSDocument` and `MSDocumentData`. So we need to check whether we have one or the other.

PR #127, opened 2018-03-13
 fix adding a path layer when accessing a shape

mathieudutour:
Resolves BohemianCoding/Sketch#17367


bomberstudios:
Change looks good, :shipit:

PR #132, opened 2018-03-14
Fix bug in Rectangle copying functionality

matt-curtis:
The former code failed to correctly copy source rectangles whose `x` == `0`, returning a Rectangle with `{ x: nan... }`. It might also be worthwhile to put some checks in place to ensure `NaN` is never possible as the values for any of the rectangle properties.


mathieudutour:
yes, good point, we could check the arguments and throw an error if it's not what we expect.

Thanks for the PR! (I just added a regression test)

PR #138, opened 2018-03-15
fix some typo in Selection.md

nichenqin:
1. `selection.clear()` does not need any parameters
1. move iterate method code & add `map` `reduce` method doc


mathieudutour:
Oh waow, I must have been tired when writing this haha. Thanks!

PR #149, opened 2018-03-21
Update sketchtool.md

jiaxianhua:
code format


mathieudutour:
Good catch! The guides are on the develop branch, in docs/guides. Could you update it there instead?

PR #157, opened 2018-03-26
Code Review: More general routine for converting rectangles between layers (#17540)

mathieudutour:
Resolves BohemianCoding/Sketch#17540 (“More general routine for converting rectangles between layers”):

> We have nice generic `convertPoint:toLayer:` for example. There should be an equivalent for converting rectangles.

Connect to BohemianCoding/Sketch#17540.


docs/api/Rectangle.md
```diff
@@ -47,6 +47,25 @@ Adjust the rectangle by scaling it. The `scaleHeight` argument can be ommited to
...
+  to: layerB,
+})
+```
+
+Each layer defines its own coordinates basis. You can change that basis from one layer to the other with `changeBasis`.
+
+| Parameters                                                 |                                                                   |
+| ---------------------------------------------------------- | ----------------------------------------------------------------- |
+| change<span class="arg-type">object</span>                 |                                                                   |
+| change.from<span class="arg-type">Layer / undefined</span> | The layer in which the rectangle's coordinates are expressed.     |
```
robintindale: is `undefined` to show the param is optional? It looks like convention is to use `required` if not optional, and nothing if the param is optional?

mathieudutour: Indeed!

mathieudutour: Does the API make sense otherwise?

docs/api/Rectangle.md
```diff
@@ -47,6 +47,25 @@ Adjust the rectangle by scaling it. The `scaleHeight` argument can be ommited to
...
+  to: layerB,
+})
+```
+
+Each layer defines its own coordinates basis. You can change that basis from one layer to the other with `changeBasis`.
+
+| Parameters                                            |                                                                   |
+| ----------------------------------------------------- | ----------------------------------------------------------------- |
+| change<span class="arg-type">object - required</span> |                                                                   |
+| change.from<span class="arg-type">Layer</span>        | The layer in which the rectangle's coordinates are expressed.     |
```
robintindale: can be `[Layer](#Layer)`

Source/dom/components/Layer.js
```diff
@@ -63,36 +63,25 @@ export class Layer extends WrappedObject {
...
   }
 
-  /**
-   * Convert a rectangle in the coordinates that this layer uses to it's parent's coordinates.
-   *
-   * @param {Rectangle} rect The rectangle to convert.
-   * @return {Rectangle} The converted rectangle expressed in the coordinate system of the parent layer.
-   */
+  // @deprecated
   localRectToParentRect(rect) {
```
robintindale: api philosophy question, as I know you've mentioned this before - is the idea to reduce the footprint of the JS api as small as possible, and leave convenience methods to user-land? In this case, it would still be cool to have `layer.parentRect` and `layer.pageRect` as that probably covers a lot of cases on what people will be trying to do, instead of them having to go and find the concepts of `basis` and get their head around that.

docs/api/Rectangle.md
```diff
@@ -47,6 +47,25 @@ Adjust the rectangle by scaling it. The `scaleHeight` argument can be ommited to
 
 Return this rectangle (useful if you want to chain the calls).
 
+## Change the coordinates basis
```
robintindale: as you asked for general API comments, `basis` is a new concept that people might be unfamiliar with. It wasn't immediately clear, but reading it obviously makes sense. Perhaps people are more used to the idea of `coordinate "space"`? or yeah `origin`? I guess there are a few generic Maths-y terms. But I don't have any real objection to basis, just a passing comment.

docs/api/Rectangle.md
```diff
@@ -47,6 +47,25 @@ Adjust the rectangle by scaling it. The `scaleHeight` argument can be ommited to
...
+
+Each layer defines its own coordinates basis. You can change that basis from one layer to the other with `changeBasis`.
+
+| Parameters                                            |                                                                   |
+| ----------------------------------------------------- | ----------------------------------------------------------------- |
+| change<span class="arg-type">object - required</span> |                                                                   |
+| change.from<span class="arg-type">Layer</span>        | The layer in which the rectangle's coordinates are expressed.     |
+| change.to<span class="arg-type">Layer</span>          | The layer in which the rectangle's coordinates will be expressed. |
+
+Both `from` and `to` can be ommited (but not at the same time) to change the basis from/to the Page coordinates.
```
robintindale: ah - I was confused until I read this :) I would structure this slightly differently if possible - it would be better if this was explicit in the parameters e.g. `default - currentPage` or something like that. I missed this note for a bit when evaluating the API, so I was left wondering what will happen if I leave them out

robintindale:
general question - is there any way to get a preview of the https://developer.sketchapp.com/reference/api/ from a branch

docs/api/Rectangle.md
mathieudutour: Interesting. The "proper" maths-y term is `basis` tho, which means "system of coordinates" (in some specific cases that happen to be the normal case 😄) :)

But maybe I can explain a bit more what is a `basis` with some links to learn more

mathieudutour:
No there is no way to preview a branch yet and since the website is in the `gh-pages` branch, I'm not sure netlify could handle it either. I'll have a look

Source/dom/components/Layer.js
mathieudutour: I added some examples to about getting the `parentRect` and `pageRect`. Hopefully, that will make it easier.

Otherwise, feel free to create a `sketch-geometry` package and I'll be happy to link to it :)

docs/api/Rectangle.md
robintindale: hehe I guess emphasis on the "generic". I either can't remember or never studied linear algebra that deeply despite doing a CS degree. But yeah `changeCoordinateSystem`, `changeCoordinateOrigin` `changeCoordinateSpace` probably aren't that snappy as a method name, despite maybe being more math-noob friendly. Anyway, was just my first reaction :)

docs/api/Rectangle.md
```diff
@@ -54,18 +54,31 @@ var newRect = rect.changeBasis({
...
-| change.to<span class="arg-type">Layer</span>          | The layer in which the rectangle's coordinates will be expressed. |
+| Parameters                                               |                                                                   |
+| -------------------------------------------------------- | ----------------------------------------------------------------- |
+| change<span class="arg-type">object - required</span>    |                                                                   |
+| change.from<span class="arg-type">[Layer](#layer)</span> | The layer in which the rectangle's coordinates are expressed.     |
+| change.to<span class="arg-type">[Layer](#layer)</span>   | The layer in which the rectangle's coordinates will be expressed. |
 
 Both `from` and `to` can be ommited (but not at the same time) to change the basis from/to the Page coordinates.
 
+If you want to explore the beautiful world of linear algebra from which the term `Basis` originates from, Khan Academy has some really nice [videos](https://www.khanacademy.org/math/linear-algebra/alternate-bases/change-of-basis/v/linear-algebra-change-of-basis-matrix).
```
robintindale: heh I think the top description is fine - this is a bit superfluous and just sort of pollutes the API

Source/dom/components/Layer.js
robintindale: yep the examples are there, was more a general question as we seem to be against these sort of convenience methods.

robintindale:
Lookin good. I didn't test any functionality though

docs/api/Rectangle.md
mathieudutour: hahaha

robintindale: 🙈

bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  `Source/dom/Rectangle.js` changed, but not its tests. That's OK as long as you're refactoring.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/components/Group.js` changed, but not its tests. That's OK as long as you're refactoring.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/components/Layer.js` changed, but not its tests. That's OK as long as you're refactoring.
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #161, opened 2018-03-28
Fix incorrect path to debug log in debugging guide

hamishmcneill:


hamishmcneill:
Closing this as the original path is correct, of course. `Plugin Developer.log` is intended for deprecation warnings.

PR #165, opened 2018-04-03
add note about entr and restarting sketch during dev

cameron:
The "in conjunction with"... section may be over the top, but it's fundamental to my workflow, and idk what the average Sketch plugin developer's familiarity with unix is, so I erred on the side of overly informative.

### DIFF
--- a/docs/guides/preferences.md
+++ b/docs/guides/preferences.md
@@ -89,7 +89,7 @@ What it doesn’t do, however, is change when a new JavaScript context is made. 
 
 If you find yourself in the latter category, of needing to restart long-running JavaScript contexts regularly during development, the unix utility [`entr`](http://entrproject.org/) may come in handy. Given a list of files on stdin, it will run a command (or restart a long-running process given `-r`) every time one of those files is modified:
 
-`find ./FB\ SketchKit.sketchplugin -name '*js' | entr -r /Applications/Sketch.app/Contents/MacOS/Sketch`
+`find /your/plugin/build/dest -name '*js' | entr -r /Applications/Sketch.app/Contents/MacOS/Sketch`
 
 #### In conjunction with webview JavaScript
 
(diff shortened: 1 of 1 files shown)


_guides/preferences.md
```diff
@@ -85,6 +85,19 @@ Please note that this setting determines whether the source of the script is rel
 
 What it doesn’t do, however, is change when a new JavaScript context is made. For long-running scripts, the same context is held in memory (it has to be — the running script is using it) until the script exits. So if you’re testing a long-running script, you will still have to find a way to stop the script, so that the context gets thrown away (that usually means relaunching Sketch, or setting `coscript.setShouldKeepAround(false)`).
 
+### Always restarting sketch after changes to plugin JavaScript
+
+If you find yourself in the latter category, of needing to restart long-running JavaScript contexts regularly during development, the unix utility `[entr](http://entrproject.org/)` may come in handy. Given a list of files on stdin, it will run a command (or restart one) every time one of those files is modified:
```
mathieudutour: I should be `[\`entr\`](http://entrproject.org/)` otherwise the link is broken.

cameron:
Comme ça?

cameron:
@mathieudutour i don't use skpm, so i'll leave that up to you 🙃but i **would** be surprised to learn that sketch knows whether or not my plugin has a webview... like, i'm using the Obj-C API to create it, so...

mathieudutour:
> I think Mathieu is referring to a plugin using skpm where we know the location of the link and we can see whether the plugin is including sketch-module-web-view or has html resources

Indeed

> This adds up to a possible idea of providing people an easy way to squash their long-running contexts (safely, as they write their own cleanup) and therefore eradicate the need to restart Sketch.

I was also thinking about something along those lines. This also touches another issue: right now Sketch is polling the long-running plugins to know when they are finished. This can cause some issues where a long-running script has finished running but hasn't been cleaned up yet and you run the same script again: it will re-use the same JSContext which is due to shut down.

So I was actually thinking about being more aggressive with fibers: when the last fiber is cleaned up, we tell Sketch to clean up the script.

That's probably for another discussion tho. Let's merge this, I'll open an issue on the skpm repo to see what we can do there

cameron:
> Mathieu is referring to a plugin using skpm...

lolo thanks for not sending a lmgtfy link 😂

> > there might be some cases where the fiber will be cleaned up by Sketch so you need to account for that
> 
> hence why we provide fiber.onCleanup. I then checked whether onCleanup is run when I used tearDownSession, but it doesn't seem to be, so that would need fixing.

We have a symptom that fits this diagnosis, so I'd be tickled to see `.onCleanup` working. A coworker says it might be in v50 beta?

Anyway. 👏

PR #168, opened 2018-04-10
Fix onOpenDocument example function

KevinGutowski:
Message toast wasn’t showing up because the dictionary structure of **context** is slightly different.


mathieudutour:
Good catch. The guide is actually in the develop branch, in docs/guides. Could you rebase your PR?

KevinGutowski:
oops yeah!

PR #179, opened 2018-04-20
tests: use npm ci

DanielRuf:
`npm ci` is much faster and is meant for CI. This PR is meant for evaluating the new ci command and testig the performance and compare the logs of CircleCI.

https://docs.npmjs.com/cli/ci
http://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -21,7 +21,8 @@ jobs:
           # fallback to using the latest cache if no exact match is found
           - v1-dependencies-
 
-      - run: npm install
+      - run: npm i -g npm@latest
+      - run: npm ci
 
       - save_cache:
           paths:
--- a/circle.yml
+++ b/circle.yml
@@ -21,7 +21,7 @@ jobs:
           # fallback to using the latest cache if no exact match is found
           - v1-dependencies-
 
-      - run: npm i -g npm@latest
+      - run: sudo npm i -g npm@latest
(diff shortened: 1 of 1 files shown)


robintindale:
I think it's slower (12s) because

> If a node_modules is already present, it will be automatically removed before `npm ci` begins its install.

from https://docs.npmjs.com/cli/ci

and we are currently caching `node_modules` between builds https://github.com/DanielRuf/SketchAPI/blob/b98f53d72efc84f2714b8fd9a60b0628f6a6e739/circle.yml#L29

which means a normal `npm install` comes out faster (7s) https://circleci.com/gh/BohemianCoding/SketchAPI/124

DanielRuf:
Makes sense. Should I close this PR then?

robintindale:
You can wait for the official response :) It may be worth it to use `npm ci` and not cache `node_modules`, as I can see it having benefits when there are actual packages to install, but it depends how often that is the case

mathieudutour:
Installing the latest npm is probably slower than installing the modules on top of a cached node_modules ;)

Let’s wait for the npm to ship with `npm ci` by default and revisit then

DanielRuf:
You are right =) Currently checking how much the improvement of this would be (pure `npm ci` step.

PR #199, opened 2018-05-22
Code Review: Improve support for Libraries in SketchAPI (#17854)

mathieudutour:
Resolves BohemianCoding/Sketch#17854

- [x] get details when a library was last updated (as shown in preferences)
- [x] enable/disable a library
- [x] get path of document if any

The following needs a block which is not possible in CocoaScript yet. So I'll revisit this when I've added it to CocoaScript (which we need anyway, more and more APIs use blocks)
- [ ] add a library from a URL

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -72,6 +72,8 @@ globals:
   MSSharedTextReference: false
   MSUserAssetLibrary: false
   MSRemoteAssetLibrary: false
+  NSSaveAsOperation: false
+  NSSaveOperation: false
 
 rules:
   ###########
--- a/CHANGELOG.json
+++ b/CHANGELOG.json
@@ -1,5 +1,8 @@
 {
   "unreleased": [
+    "[New] Add `path` property to Document",
+    "[New] Add `lastModifiedAt` property to Library",
+    "[New] Disable/Enable a Library by setting its `enabled` property",
     "[Fixed] Setting an image override on a symbol instance",
     "[Fixed] Saving a document without specifying a path would fail"
   ],
--- a/Source/dom/components/Document.js
+++ b/Source/dom/components/Document.js
@@ -1,10 +1,32 @@
-import { WrappedObject, DefinedPropertiesKey } from '../WrappedObject'
-import { Page } from './Page'
-import { Selection } from '../Selection'
-import { toArray, getURLFromPath, getDocumentData } from '../utils'
-import { wrapObject } from '../wrapNativeObject'
-import { Types } from '../enums'
-import { Factory } from '../Factory'
+import {
+  WrappedObject,
+  DefinedPropertiesKey
+} from '../WrappedObject'
+import {
+  Page
+} from './Page'
+import {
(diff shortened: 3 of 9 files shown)


bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Changes were made to package.json, but not to package-lock.json - <i>Perhaps you need to run `npm run install`?</i>
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

ivanortegaalba:
Hi @mathieudutour! Thanks for this improvement! It's really amazing!

About the last point, the import library from URL, I did something like:
```
/**
 * Add a Sketch file as Library for a given URL.
 * @param {String} libUrl FileURL where the Sketch file is located.
 * @returns {Library} The library instance for the given URL
 */
function getLibraryForDocumentAtURL(libUrl) {
    const STATUS = {
        0: 'ok',
        1: 'the library has already been added',
        2: 'the document is not in the new JSON format',
        3: 'there was a problem reading the asset library file',
    };
    const libraryController = AppController.sharedInstance().librariesController();

    // check if the library is already imported and remove it
    const existingLibraries = libraryController.libraries();

    for (let i = 0; i < existingLibraries.count(); i += 1) {
        const existingLibrary = existingLibraries.objectAtIndex(i);
        const location = existingLibrary.locationOnDisk();

        if (location && location.isEqual(libUrl)) {
            return fromNative(existingLibrary);
        }
    }

    // otherwise add it
    const status = libraryController.addAssetLibraryAtURL(libUrl);

    if (status !== 0) {
        throw new Error(`Error while adding the library: ${STATUS[status]}.`);
    }

    const lib = libraryController.userLibraries().firstObject();

    if (!lib) {
        throw new Error('Could not find the added library');
    }

    // refresh the UI
    libraryController.notifyLibraryChange(lib);

    return Library.fromNative(lib);
}

```

It's a little modification of `getLibraryForDocumentAtPath`. Is it not valid this way?

mathieudutour:
hum a remote library is defined by an appcast URL, not the sketch file itself. So I don't think that will work, does it? It should look more like this: https://github.com/BohemianCoding/SketchAPI/pull/199/files#diff-dd8ceacc64098749be04d8a9a8b6d848R81 but I didn't have time to finish it yet

ivanortegaalba:
Oh ok! This means for Libraries in a remote server. I was using the previous implementation to add a collection of symbols in a sketch file, keeping this Sketch file in the plugin asset folder. So, doing this: 
```js
const libraryURL = context.plugin.urlForResourceNamed('NR-UI.sketch');
getLibraryForDocumentAtURL(libraryURL);
```
I add the file as a library. 

For me, this way was easier than using the path to add the file as a library.
It's very useful to distribute a collection of symbols (even Design Systems) through a plugin.

Sorry about the misunderstanding. Thank you!

PR #200, opened 2018-05-23
fix indents

pravdomil:
see diff


mathieudutour:
Thanks! Could you create the PR onto the `develop` branch (the guides are under `docs/guides/`?

pravdomil:
#201

Issue #224, opened 2018-06-29
Setting document settings with cocoa NSNumber results in {}


WCByrne:
I have some NSValue and NSNumber values that I would like to store in the document settings. Attempting to pass these values to `setDocumentSettingsForKey()` results in empty objects `{}` being stored. To be precise when printing the className of the value it is NSCFNumber.

Do these object values need to be converted somehow before being passed to the settings?

mathieudutour:
For NSString, you can wrap them in a string: `let string = String(anNSString)`

I do agree that NSNumber, NSString, NSDictionnary and NSArray should be parsed into their JS counterpart. You could use the [`replacer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#The_replacer_parameter) argument of the `JSON.stringify` function.

There are a bunch of functions in the [`util`](https://github.com/skpm/util) that already takes care of mapping native objects to JS

WCByrne:
I thought I had tried `String(ns)` but apparently not. That does work. I'll take a look at a generic solution for my needs and try to come up with something that could make it's way upstream. 

I haven't had a chance to look thoroughly yet, Is `stringify` used throughout the API, or just a few places (like settings)?

mathieudutour:
just the settings, because a setting ultimately is a string

WCByrne:
Not 100% sure how you are managing contributions but I used this as an excuse to get familiar with the   process: https://github.com/WCByrne/SketchAPI/tree/coerce-ns-settings-values

mathieudutour:
We use pull request. Open one and I'll review it :)

PR #225, opened 2018-06-30
Improve stringification of NSObjects 

WCByrne:
#224 

Implicitly convert some common NSObjects to native JS objects to improve stringification when storing settings.

### DIFF
--- a/Source/settings/Settings.js
+++ b/Source/settings/Settings.js
@@ -1,6 +1,5 @@
+import util from 'util'
 import { isWrappedObject, getDocumentData } from '../dom/utils'
-
-const util = require('util')
 
 function getPluginIdentifier() {
   if (!__command.pluginBundle()) {
@@ -118,7 +117,6 @@ export function documentSettingForKey(document, key) {
 }
 
 export function setDocumentSettingForKey(document, key, value) {
-  log(util)
   const documentData = getDocumentData(document)
   const stringifiedValue = JSON.stringify(value, (k, v) => util.toJSObject(v))
   __command.setValue_forKey_onDocument(stringifiedValue, key, documentData)
--- a/Source/settings/Settings.js
+++ b/Source/settings/Settings.js
@@ -12,30 +12,6 @@ function getPluginIdentifier() {
 }
 
 /**
- * Coerce common NSObjects to their JS counterparts
- * @param value Any value
- *
- * Converts NSDictionary, NSArray, NSStirng, and NSNumber to
- * native JS equivilents.
- */
-function coerceNSObject(value) {
-  if (value) {
-    if (util.isObject(value)) {
-      return util.toObject(value)
-    } else if (util.isArray(value)) {
-      return util.toArray(value)
-    } else if (util.isString(value)) {
-      return String(value)
-    } else if (util.isNumber(value)) {
-      return Number(value)
-    } else if (util.isBoolean && util.isBoolean(value)) {
-      return Boolean(Number(value))
-    }
-  }
-  return value
-}
-
-/**
  * Return the value of a global setting for a given key.
  * @param key The setting to look up.
  * @return The setting value.
@@ -64,7 +40,7 @@ export function globalSettingForKey(key) {
  */
 export function setGlobalSettingForKey(key, value) {
   const store = NSUserDefaults.standardUserDefaults()
-  const stringifiedValue = JSON.stringify(value)
+  const stringifiedValue = JSON.stringify(value, (k, v) => util.toJSObject(v))
   if (!stringifiedValue) {
     store.removeObjectForKey(key)
   } else {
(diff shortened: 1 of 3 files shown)


Source/settings/Settings.js
```diff
@@ -9,6 +9,98 @@ function getPluginIdentifier() {
   return __command.pluginBundle().identifier()
 }
 
+// check if the argument is a native sketch object
```
mathieudutour: so I didn't properly documented yet so it's my bad but there are some packages shipping with sketch and we should take advantage of them instead of rewriting that:

```
const util = require('util')

function coerceNSObject(value) {
  if (value) {
    if (util. isObject(value)) {
      return util.toObject(value)
    } else if (util.isArray(value)) {
      return util.toArray(value)
    } else if (util.isString(value)) {
      return String(value)
    } else if (util.isNumber(value)) {
      return Number(value)
    }
  }
  return value
}
```

WCByrne: I agree this could be nice to expose to plugins. I'm not sure which util package you are referring to. I'm unable to build with `require('util')`.

Also, forgive any obvious questions. I'm still fitting my JS hat back on after a long hiatus.

mathieudutour: indeed, you need to tell webpack that `util` is an external dependency: In the webpack.config.js file, you need to add
```js
const CORE_MODULES = Object.keys(
  require('./core-modules/package.json').dependencies
).map(k => k.replace('skpm/', ''))

...
const config = {
  externals: [
    (context, request, callback) => {
      // core modules shipped in Sketch
      if (CORE_MODULES.indexOf(request) !== -1) {
        return callback(null, `commonjs ${request}`)
      }
      return callback()
    },
  ],
...
```

WCByrne: 👍 Got that working, thanks. 

As for using that, there are a couple issues I'm fining with the functions in the util conversions. 

1. Object and Array don't do deep conversions
2. NSNumber with a bool is type NSCFBoolean which is not handled

mathieudutour: > Object and Array don't do deep conversions

Indeed, we'd still need that. I though the replacer in the JSON.stringify was called recursively? Maybe it's only for objects that aren't supplied by a previous replacer.

> NSNumber with a bool is type NSCFBoolean which is not handled

Good catch, I'll change the util package to handle that case.
The coerce method will need another case:
```js
 else if (util.isBoolean(value)) {
      return Boolean(Number(value))
    }
```

WCByrne:
Implemented using util but it's not seeing some of the methods there (i.e. toObject). printing `util` I see all of the `is***()` methods but none of the `to***()` ones, not sure why this is.

Also, moving the remaining `coerceNSObject` to `util` definitely seems like the best option now to keep all conversion logic together. Probably with a name more consistent with the other `to***()`.

mathieudutour:
yes that sounds good. The package is here: https://github.com/skpm/util

WCByrne:
The changes here are pretty trivial now but depend on the updated `skpm/util` package. I'm not quite sure the best way to go about testing the full stack of changes. I see `skpm/util` in `core-modules/package.json` but don't see that being used by the dev API. Any tips here?

mathieudutour:
I don't really have a good way unfortunately. What would be nice is to:
- add some tests in the `skpm/util` repo (and you can run the tests using `npm run test` there)
- update the version of `skpm/util` in the `core-modules/package.json`
- then I'll create a PR on the Sketch repo so that the tests will run with the new `skpm/util` package

WCByrne:
Sorry, I got a bit sidetracked from this for a bit. I was just adding some tests to util but am getting some odd behavior. If I remember correctly this was working for document settings but the tests aren't returning what I was expecting for nested objects. Not sure if it is a problem though. 

I left another comment in that PR https://github.com/skpm/util/pull/4

mathieudutour:
so looking at the code, it's expected that nested objects aren't coerced since the function is not recursive.

It's already a goog step forward but it might be worth fixing it for good?

WCByrne:
Oh, right. It is the replacer that is recursive 👍

bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/settings/Settings.js` changed, but not its doc. That's OK as long as you're refactoring.
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

Source/settings/Settings.js
```diff
@@ -115,6 +118,8 @@ export function documentSettingForKey(document, key) {
 }
 
 export function setDocumentSettingForKey(document, key, value) {
+  log(util)
```
mathieudutour: extra log

Source/settings/Settings.js
```diff
@@ -1,5 +1,7 @@
 import { isWrappedObject, getDocumentData } from '../dom/utils'
 
+const util = require('util')
```
mathieudutour: can we use `import from` to be consistent?

PR #232, opened 2018-07-21
update "Running your plugin"

tsanthosh:
add `npm install` step before building the plugin with `npm run build`.


mathieudutour:
`skpm create` should install the dependencies for you already so you don’t have to run `npm install`. 

Did you find a case where that’s not happening?

tsanthosh:
Hmm.. You are right it does work. Don't know why it did not work for me the first time. Thanks for your time.

PR #234, opened 2018-07-23
Library.getRemoteLibraryWithRSS: check for NSNull errors

romannurik:
The "if (err)" check was returning true even when err is an NSNull


mathieudutour:
Ah I actually have the same fix here already: https://github.com/BohemianCoding/SketchAPI/pull/233/files#diff-dd8ceacc64098749be04d8a9a8b6d848L101

romannurik:
Awesome!

PR #236, opened 2018-07-24
Code Review: Add support for Shared styles in the SketchAPI (#19568)

### LINKED ISSUE #118: Shared layer/text styles support
Is there any way to access/manipulate layer and text styles? Is it just missing in the documentation, or it's something that'll be provided in the future?

### LINKED ISSUE #237: Bug: getLibraries() not working when there is a missing library
scenario to reproduce:
 
add a sketch file as library, in finder move that library to a different folder.

Run the following:
```
var libraries = require('sketch/dom').getLibraries();
console.log(libraries);
```

output is:
```
TypeError: null is not an object (evaluating 'e.timeIntervalSince1970')
line: 1
sourceURL: /Applications/Sketch.app/Contents/Resources/SketchAPI_dom.js
column: 64745
```

mathieudutour:
Resolves BohemianCoding/Sketch#19568

fix #237, fix #118


docs/api/fromNative.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: fromNative
-order: 405
-section: Utils
+order: 211
+section: models
```
mathieudutour: I'm not really satisfied with the name of this section but it will do for now

Source/dom/models/Document.js
```diff
@@ -134,6 +134,18 @@ export class Document extends WrappedObject {
     return toArray(documentData.allSymbols()).map(wrapObject)
   }
 
+  getSharedLayerStyles() {
+    const documentData = this._getMSDocumentData()
+    return toArray(documentData.layerStyles().sharedStyles()).map(wrapObject)
```
opsGavin: This will only return styles local to the document. There's also an `allLayerStyles` (and equivalent `allTextStyles`) method which will include any foreign styles used within the document. This is maybe be design, but I thought I'd point it out just in case.

Source/dom/models/Library.js
```diff
@@ -119,10 +119,30 @@ export class Library extends WrappedObject {
...
+          wrapObject(document).sketchObject
+        )
+        break
+      case ImportableObjectType.TextStyle:
+        provider = MSSharedTextStyleProvider.alloc().initWithDocument(
+          wrapObject(document).sketchObject
+        )
+        break
+      default:
+        provider = MSForeignObjectProvider.alloc().initWithDocument(
```
opsGavin: We might be better throwing an error or just returning nil here?

Source/dom/models/Override.js
```diff
@@ -14,7 +14,7 @@ Factory.registerClass(Override, MSAvailableOverride)
 
 Override.define('path', {
   get() {
-    return String(this._object.overridePoint().layerIDPath())
+    return String(this._object.overridePoint().path())
```
opsGavin: Sorry - I changed that. I thought I'd updated the API stuff, but I guess not...

Source/dom/style/Style.js
```diff
@@ -181,3 +199,56 @@ Style.define('innerShadows', {
...
+        const currentSharedObject = container.sharedStyleWithID(currentSharedId)
+        currentSharedObject.unregisterInstance(this._object)
+      }
+    } else {
+      const sharedStyle = container.sharedStyleWithID(newId)
+      if (!sharedStyle) {
+        throw new Error('Seems like this shared style does not exists')
+      }
+      container.registerInstance_withSharedStyle(this._object, sharedStyle)
+    }
```
opsGavin: The styles are still in the shared containers but they aren't nice, and the plan is to gradually move away from them until we get to the point where we can just rip them out. A better way to do this would be to use the `sharedStyle` property defined in `MSLayer+SharedStyles.h`.

Source/dom/style/Style.js
```diff
@@ -181,3 +199,56 @@ Style.define('innerShadows', {
...
+  get() {
+    const { sharedStyleId } = this
+    if (!sharedStyleId) {
+      return null
+    }
+    const documentData = this._object.documentData()
+    const container = documentData.sharedObjectContainerOfType(
+      this._object.type()
+    )
+    return wrapObject(container.sharedStyleWithID(sharedStyleId))
```
opsGavin: I'm not entirely sure I'm following what kind of object this code is wrapping, but as above we want to avoid accessing the `sharedObjectContainers`. But I suspect you could again use `MSLayer+SharedStyles.h` and just return  `_this.sharedStyle`

Source/dom/models/Override.js
mathieudutour: I'm actually not sure how it passed the tests. They were failing for me /shrug

Source/dom/style/Style.js
mathieudutour: hum I don't have access to the Layer here. And worse, the style might not be inserted in a Layer but might already reference a shared style.

PR #257, opened 2018-10-02
Code Review: SharedStyle are broken in SketchAPI (#21115)

mathieudutour:
resolves BohemianCoding/Sketch#21115

### DIFF
--- a/Source/dom/layers/StyledLayer.js
+++ b/Source/dom/layers/StyledLayer.js
@@ -16,12 +16,20 @@ Factory.registerClass(StyledLayer, MSImmutableStyledLayer)
 StyledLayer.define('style', {
   get() {
     const style = Style.fromNative(this._object.style())
+    style._parentLayer = this
     return style
   },
   set(style) {
     if (this.isImmutable()) {
       return
     }
+    if (style && style._sharedStyle) {
+      this._object.setSharedStyleID(style._sharedStyle.id)
+    }
+    if (style && style.sharedStyleId) {
+      this._object.setSharedStyleID(style.sharedStyleId)
+    }
+
     if (isNativeObject(style)) {
       this._object.style = style
     } else if (!style || !style.sketchObject) {
--- a/Source/dom/layers/StyledLayer.js
+++ b/Source/dom/layers/StyledLayer.js
@@ -3,6 +3,7 @@ import { Factory } from '../Factory'
 import { Layer } from './Layer'
 import { Style } from '../style/Style'
 import { isNativeObject } from '../utils'
+import { Types } from '../enums'
 
 /**
  * Represents a layer with style.
@@ -23,6 +24,15 @@ StyledLayer.define('style', {
     if (this.isImmutable()) {
       return
     }
+
+    // handle passing a shared style directly
+    if (style && style.type === Types.SharedStyle) {
+      this._object.setSharedStyleID(style.id)
+      return
+    }
+
+    // if a style has a shared style id, we need to set it first
+    // so that the layer is linked to the shared style
     if (style && style._sharedStyle) {
       this._object.setSharedStyleID(style._sharedStyle.id)
     }
@@ -30,6 +40,7 @@ StyledLayer.define('style', {
       this._object.setSharedStyleID(style.sharedStyleId)
     }
 
+    // we can then actually set the style
     if (isNativeObject(style)) {
       this._object.style = style
     } else if (!style || !style.sketchObject) {
--- a/Source/dom/layers/StyledLayer.js
+++ b/Source/dom/layers/StyledLayer.js
@@ -38,7 +38,31 @@ StyledLayer.define('style', {
   },
 })
 
+StyledLayer.define('sharedStyleId', {
+  get() {
+    const nativeSharedStyle = this._object.sharedStyleID()
+    if (!nativeSharedStyle) {
+      return null
+    }
+    return String(nativeSharedStyle)
+  },
+  set(sharedStyleId) {
+    if (this.isImmutable()) {
+      return
+    }
+
+    if (!sharedStyleId) {
+      this._object.setSharedStyleID(null)
+      return
+    }
+
+    this._object.setSharedStyleID(sharedStyleId)
+  },
+})
+
 StyledLayer.define('sharedStyle', {
+  enumerable: false,
+  exportable: false,
   get() {
     if (!this._object.sharedObject) {
       return null
@@ -61,8 +85,5 @@ StyledLayer.define('sharedStyle', {
 
     const nativeSharedStyle = wrapObject(sharedStyle)
     this._object.setSharedStyleID(nativeSharedStyle.id)
-    this._object
-      .style()
-      .syncWithTemplateInstance(nativeSharedStyle.style.sketchObject)
   },
 })
--- a/Source/dom/layers/StyledLayer.js
+++ b/Source/dom/layers/StyledLayer.js
@@ -3,7 +3,8 @@ import { Factory } from '../Factory'
 import { Layer } from './Layer'
 import { Style } from '../style/Style'
 import { isNativeObject } from '../utils'
-import { Types } from '../enums'
+import { SharedStyle } from '../models/SharedStyle'
+import { wrapObject } from '../wrapNativeObject'
 
 /**
  * Represents a layer with style.
@@ -16,31 +17,15 @@ Factory.registerClass(StyledLayer, MSImmutableStyledLayer)
 
 StyledLayer.define('style', {
   get() {
-    const style = Style.fromNative(this._object.style())
-    style._parentLayer = this
-    return style
+    return Style.fromNative(this._object.style())
   },
   set(style) {
     if (this.isImmutable()) {
       return
     }
 
-    // handle passing a shared style directly
-    if (style && style.type === Types.SharedStyle) {
-      this._object.setSharedStyleID(style.id)
-      return
-    }
-
-    // if a style has a shared style id, we need to set it first
-    // so that the layer is linked to the shared style
-    if (style && style._sharedStyle) {
-      this._object.setSharedStyleID(style._sharedStyle.id)
-    }
-    if (style && style.sharedStyleId) {
-      this._object.setSharedStyleID(style.sharedStyleId)
-    }
-
     // we can then actually set the style
+    let nativeStyle
     if (isNativeObject(style)) {
       this._object.style = style
     } else if (!style || !style.sketchObject) {
@@ -48,5 +33,36 @@ StyledLayer.define('style', {
     } else {
       this._object.style = style.sketchObject
     }
+
+    this._object.style = nativeStyle.copy()
+  },
+})
+
+StyledLayer.define('sharedStyle', {
+  get() {
+    if (!this._object.sharedObject) {
+      return null
+    }
+    const nativeSharedStyle = this._object.sharedObject()
+    if (!nativeSharedStyle) {
+      return null
+    }
+    return SharedStyle.fromNative(nativeSharedStyle)
+  },
+  set(sharedStyle) {
+    if (this.isImmutable()) {
+      return
+    }
+
+    if (!sharedStyle) {
+      this._object.setSharedStyleID(null)
+      return
+    }
+
+    const nativeSharedStyle = wrapObject(sharedStyle)
+    this._object.setSharedStyleID(nativeSharedStyle.id)
+    this._object
+      .style()
+      .syncWithTemplateInstance(nativeSharedStyle.style.sketchObject)
   },
 })
--- a/Source/dom/models/SharedStyle.js
+++ b/Source/dom/models/SharedStyle.js
@@ -3,7 +3,6 @@ import { Types } from '../enums'
 import { Factory } from '../Factory'
 import { wrapObject } from '../wrapNativeObject'
 import { toArray } from '../utils'
-import { Style } from '../style/Style'
 
 const SharedStyleTypeMap = {
   1: 'Layer',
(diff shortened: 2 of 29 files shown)


bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/WrappedObject.js` changed, but not its tests. That's OK as long as you're refactoring.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/layers/StyledLayer.js` changed, but not its tests. That's OK as long as you're refactoring.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/style/Style.js` changed, but not its tests. That's OK as long as you're refactoring.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/layers/StyledLayer.js` changed, but not its doc. That's OK as long as you're refactoring.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/models/SharedStyle.js` changed, but not its doc. That's OK as long as you're refactoring.
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

Source/dom/style/Style.js
```diff
@@ -203,37 +232,43 @@ Style.define('innerShadows', {
 
 Style.define('sharedStyleId', {
   get() {
-    if (this._object.sharedObjectID()) {
-      return String(this._object.sharedObjectID())
+    const parentLayer = this._getNativeParentLayer()
+    if (parentLayer && parentLayer.sharedStyleID) {
```
opsGavin: I'm not sure, but this could be a little dangerous. I think if the style is the one owned by the shared style that'll be the parent and it won't implement `sharedStyleID`. We could fix this obj-C side by adding a `sharedStyleID` getter to `MSSharedStyle` which just returns the `objectID`.

I think it would be better not to have this method at all TBH, but maybe that's going to cause a lot of hurt?

Source/dom/layers/StyledLayer.js
```diff
@@ -16,12 +16,20 @@ Factory.registerClass(StyledLayer, MSImmutableStyledLayer)
...
   set(style) {
     if (this.isImmutable()) {
       return
     }
+    if (style && style._sharedStyle) {
+      this._object.setSharedStyleID(style._sharedStyle.id)
+    }
+    if (style && style.sharedStyleId) {
+      this._object.setSharedStyleID(style.sharedStyleId)
+    }
```
opsGavin: It would probably be worth having a `sharedStyle` property on this object as well? I'm not sure `style._sharedStyle` would be 100% reliable so having that as a preferred way to set it would be good.

mathieudutour: so I added the ability to set a SharedStyle directly:
```js
layer.style = sharedStyle
```

I'm a bit reluctant to add a `sharedStyle` property as the sentence we want to enable is:
- "set the style of the layer to be the same as the shared style" -> `layer.style = sharedStyle`

Source/dom/style/Style.js
mathieudutour: So I really like the idea of having the same API for symbols and shared styles:
- SymbolMaster -> SharedStyle
- SymbolInstance -> Style
- SymbolMaster.newInstance() -> SharedStyle.newInstance()
- SymbolInstance.symbolId -> Style.sharedStyleId
- etc.

opsGavin: > The signature of parentLayer is MSStyledLayer

```
- (MSStyledLayer *)parentLayer {
  id parent = self.parentObject;
  if ([parent isKindOfClass:[MSStyledLayer class]])
    return parent;
  else
    return nil;
}
```

Ah - yes. You're right - that'll be safe.

> So I really like the idea of having the same API for symbols and shared styles

I agree but I think I see SymbolInstance mapping to SharedStyle rather than Style.
instance.symbolMaster -> layer.sharedStyle

mathieudutour: but a layer is not an instance of a sharedStyle 🤔

opsGavin: 🤦‍♂️ Sorry - what I meant to say was:

I agree but I think I see SymbolInstance mapping to ~~SharedStyle~~ StyledLayer rather than Style.
instance.symbolMaster -> layer.sharedStyle

mathieudutour: That's the sketch model yes but without knowing the internals, I don't think that's the most intuitive interface. You cannot:
- create a layer from a shared style, there is too much missing from it.
- check if a layer is the same as a SharedStyle (you could be, but the mental model would be really weird to me)
- sync a layer with a sharedStyle
- etc.

opsGavin: I can see where you're coming from, but I think there are multiple ways to look at the analogy and sooner or later one of them will break down.

* You create a new symbol instance - you create a new styled layer.
* You link the symbol instance to a symbol master - you link the styled layer to a shared style.
* You lookup the symbol master by ID in the document - you lookup the shared style by ID in the document.
* You can import symbol masters as foreign objects - you can import shared styles as foreign objects,
* You modify the layers in the symbol master - you modify the fills etc in the shared style
* A symbol master is an artboard, but an artboard isn't necessarily a symbol master - a shared style has a style, but a style doesn't necessarily belong to a shared style.

But that breaks down because you can't modify a symbol instance to give it's layers different properties the way you can modify a layer's local style. You can do stuff with overrides, but you couldn't, say, add a layer to a local copy of the master without also adding it to the master.

I suppose my worry about this really is that in Sketch setting `layer.style` and `layer.sharedStyle` do two quite different things. The former modifies the local version of the layer's style while the later links the layer to the shared style and syncs it to match. I also worry that by trying to make them the same thing a user could potentially see different behaviour setting `style` where they wouldn't expect it. For instance (at least in the obj-C world) if you did this:

`layer.style = otherLayer.style`

layer would end up referencing the shared style, whereas if you did this:

`layer.style = otherLayer.style.copy`

It wouldn't - because the copy wouldn't have a parent.

Maybe there's no way we'd end up with this sort of behaviour in the API as it stands, but if we did that would be bad. I think this design makes that a possibility whereas with two distinct methods it wouldn't be.

mathieudutour: > You create a new symbol instance - you create a new styled layer

You can't actually create a new styled layer, it's an abstract class so the behaviour of `sharedStyle.newInstance()` would be undefined.

I agree with all your other points. The 2 analogies are valid but I believe the SharedStyle -> Style makes more sense without knowing how the internals of Sketch work. Both analogies are likely to break anyway at some point but this JS API is here as a buffer.

Regarding `layer.style = otherLayer.style` vs `layer.style = otherLayer.style.copy`, I'm not sure I understand your point but setting `layer.style = otherLayer.style` will end up referencing the shared style in this API. And that's what I would expect. There is no `copy` in JS.

Source/dom/layers/StyledLayer.js
WCByrne: As a use case example that is a little confusing to me; Would it be possible to set a Style, associated with a SharedStyle without linking it? I would expect setting a Style would _not_ link, while setting a SharedStyle would.

mathieudutour: yes:
```
const style = sharedStyle.newInstance() // create a new style instance linked to the shared style
style.sharedStyle = undefined // unlink the style
layer.style = style
```

mathieudutour: could you write a small code example of what it would look like? I'm struggling to visualize it

WCByrne: I think the tricky part to understand without knowing the internals is that, in a sense,  a Style can "represent" a SharedStyle (if it has a sharedStyleId).

Without realizing that I would expect the following
```javascript
const aSharedStyle = // a SharedStyle
layer.sharedStyle = aSharedStyle // sets and links the style
layer.style = aSharedStyle.style // just sets the style w/out linking
// and maybe
layer.sharedStyle = aSharedStyle.style // sets and links the style
```
In other words, to link you would have to interact with the sharedStyle property. But again, that doesn't make as much sense knowing the internals

A possible example of confusion right now:
```javascript
const styleA = layerA.style // not shared
const styleB = layerB.style // shared style
layer.style = styleA // just sets the style 
layer.style = styleB // sets and links the style
```
Most of the time this is probably the intended behavior which is why I say it seems like a good option, so long as it is well documented.

mathieudutour: ok, I resigned myself to limit the analogy to just SymbolMaster/SharedStyle. There won't be any analogy with SymbolInstance even tho there is a `SharedStyle.getAllInstances()` which returns Styles.

So now there is a `layer.sharedStyleId` property (and a convenient `layer.sharedStyle` as well). Setting it does NOT change the style of the layer. To sync the style of a layer with a shared style, you will need to do `layer.style.syncWithSharedStyle(layer.sharedStyle)`.

Source/dom/models/SharedStyle.js
```diff
@@ -22,37 +22,30 @@ export class SharedStyle extends WrappedObject {
...
     )
 
     container.addSharedObject(sharedStyle._object)
-    wrappedInstance.sharedStyleId = sharedStyle.id
 
     return sharedStyle
   }
 
-  // Returns a new Style instance linked to this SharedStyle, ready for adding to a layer
-  createNewInstance() {
```
WCByrne: Won't this be a breaking change?

Source/dom/layers/StyledLayer.js
WCByrne: Do **you** think this is better? I'm afraid I should have placed more emphasis on "Most of the time this is probably intended". I hate to say this after you've done the work but seeing the alternative, the "magic" approach does seem cleaner, you can set a Style (w/ or w/out a sharedID) or SharedStyle, and the style you set will define how it is attached.

I assume linking and applying is the more common goal, rather than just linking. Just linking could still be done by applying the sharedStyleId on the style though right?

Source/dom/models/SharedStyle.js
mathieudutour: It never worked so not really 😁

Source/dom/layers/StyledLayer.js
mathieudutour: I’m not too sure either way. Some cases makes more sense before (as in the analogy between SymbolInstance and Style made some sense) but at the same time I think it’s more explicit now and explicit always win

Source/dom/layers/StyledLayer.js
```diff
@@ -15,19 +17,73 @@ Factory.registerClass(StyledLayer, MSImmutableStyledLayer)
...
+      return
+    }
+
+    if (!sharedStyle) {
+      this._object.setSharedStyleID(null)
+      return
+    }
+
+    const nativeSharedStyle = wrapObject(sharedStyle)
+    this._object.setSharedStyleID(nativeSharedStyle.id)
```
WCByrne: You said setting the sharedStyle won't update the style but will it set the sharedStyleID on the layer and the sharedStyleID of the style? Hard to tell without what is happening internally when this is called. In other words, is the following always true after setting any of the (shared)style properties? `layer.style.sharedId == layer.sharedStyleID == layer.sharedStyle`

mathieudutour: A style doesn’t have sharedStyleId anymore. It is not linked to SharedStyle in any way

WCByrne: Any reason to not apply the style here? I can't think of a reason you wouldn't want to do both. If not, any thoughts on adding a helper like Style used to have? So you could replace:
```
layer.sharedStyle = aSharedStyle
layer.style.syncWithSharedStyle(layer.sharedStyle)
```
with
```
layer.sharedStyle = aSharedStyle
layer.syncWithSharedStyle()
```

docs/api/SharedStyle.md
```diff
@@ -47,7 +47,7 @@ Create a new Shared Style with a specific name in a specific Document.
 
 A new SharedStyle
 
-## Create a new Instance
```
WCByrne: Isn't this gone?

docs/api/Group.md
```diff
@@ -18,6 +18,7 @@ A group of layers. It is also an instance of [Layer](#layer) so all the methods
 | frame<span class="arg-type">[Rectangle](#rectangle)</span> | The frame of the Group. This is given in coordinates that are local to the parent of the layer. |
 | flow<span class="arg-type">[Flow](#flow)</span>            | The prototyping action associated with the Group.                                               |
 | style<span class="arg-type">[Style](#style)</span>         | The style of the Group.                                                                         |
+| sharedStyleId<span class="arg-type">string / null</span>   | The ID of the [SharedStyle](#sharedstyle) this Group is linked to if any.                       |
```
WCByrne: Should this have a note about having to call `layer.style.syncWithSharedStyle(sharedStyle)` somewhere? Would be nice to not have to repeat these across all StyleLayer subclasses 😕

WCByrne: Anything about `sharedStyle` or just the `sharedStyleId`??

mathieudutour: > note about having to call layer.style.syncWithSharedStyle(sharedStyle) somewhere

Yeah probably, not sure where. Maybe in the Style doc 🤔 

> Anything about sharedStyle or just the sharedStyleId??

That's a good question. `sharedStyle` is not enumerable, meaning that is you do `console.log(layer)`, it won't show `sharedStyle`. There is also the `parent` property and a few others. I haven't documented any of them yet, we probably need another section for each document.

Feel free to send a PR, we wanted to merge it quickly to include it in Sketch 52.1

docs/api/SharedStyle.md
mathieudutour: that's why it's removed?

WCByrne: 👍 Oops, I must have been looking at an earlier commit before it was removed.

PR #258, opened 2018-10-03
Fixed typo

sonburn:
Should be "new Page" not "new Artboard" since providing instructions for creating a page.


mathieudutour:
Thanks!

PR #272, opened 2018-10-24
fix fragments

### LINKED ISSUE #135: Text API: line fragments are always full-width
Is it possible for the `Text.fragments` to report more precise rectangles? While it's great to be able to get the individual rectangles of each line, the width of the rectangles is always equal to that of the text layer. Since it's not difficult to get the width of the layer itself, it would be more useful if it returned the actual width of the line as fitted to its characters.

### LINKED ISSUE #270: text.fragments is throwing an error
`text.fragments` is throwing an error since 52:

```ObjC method new requires 0 arguments, but JavaScript passed 1 argument```

### PR DESCRIPTION
**Pull request by mathieudutour**
fix #270, fix #174, fix #135

### DIFF
--- a/Source/dom/layers/__tests__/Text.test.js
+++ b/Source/dom/layers/__tests__/Text.test.js
@@ -100,11 +100,38 @@ test('should return the fragments of a text layer', () => {
 
   // https://github.com/BohemianCoding/SketchAPI/issues/144
   text = new Text({
-    text: 'Test\nHello\n123\n',
+    text: 'Test\nHello\n123\no',
   })
   text.adjustToFit()
   // eslint-disable-next-line
   fragments = text.fragments
 
   expect(fragments.length).toBe(4)
+  expect(fragments[0].baselineOffset).toBe(3)
+  expect(Number(fragments[0].range.location)).toBe(0)
+  expect(Number(fragments[0].range.length)).toBe(5)
+  expect(fragments[0].rect.toJSON()).toEqual({
+    x: 0,
+    y: 0,
+    width: 10000000,
+    height: 14,
+  })
+  expect(fragments[1].baselineOffset).toBe(3)
+  expect(Number(fragments[1].range.location)).toBe(5)
+  expect(Number(fragments[1].range.length)).toBe(6)
+  expect(fragments[1].rect.toJSON()).toEqual({
+    x: 0,
+    y: 14,
+    width: 10000000,
+    height: 14,
+  })
+  expect(fragments[2].baselineOffset).toBe(3)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Source/dom/layers/__tests__/Text.test.js [hunk 1]
```diff
@@ -78,3 +78,60 @@ test('should fix the width', () => {
...
+  let { fragments } = text
+
+  expect(fragments.length).toBe(1)
+  expect(fragments[0].baselineOffset).toBe(3)
+  expect(Number(fragments[0].range.location)).toBe(0)
+  expect(Number(fragments[0].range.length)).toBe(4)
+  expect(fragments[0].rect.toJSON()).toEqual({
+    x: 0,
+    y: 0,
+    width: 10000000,
```
bomberstudios: This looks like a magical number, not sure what's going on here.

Source/dom/layers/__tests__/Text.test.js
```diff
@@ -78,3 +78,60 @@ test('should fix the width', () => {
 
   expect(text.fixedWidth).toBe(true)
 })
+
+test('should return the fragments of a text layer', () => {
+  let text = new Text({
+    text: 'blah',
+  })
+  text.adjustToFit()
```
bomberstudios: Totally unrelated to this PR, but I wonder if we should make the `text.adjustToFit()` bit part of the initialisation for `Text` objects? Or is that something that we don't want to do all the time?

Also, would it be nice / useful to support a simpler method of creating a Text layer? Something like `let text = new Text('blah')`?

Source/dom/layers/__tests__/Text.test.js
mathieudutour: Found it! Not sure it ever worked properly...

Source/dom/layers/__tests__/Text.test.js
mathieudutour: > if we should make the `text.adjustToFit()` bit part of the initialisation for Text objects? Or is that something that we don't want to do all the time?

I guess when the `frame` is not specified we could do it automatically. If it's specified, I don't think we should.

> Something like let `text = new Text('blah')`?

That'd be really easy to do but I'm not convinced it's worth it: it creates an asymmetry between layers, harder to document, increase the API surface, etc.. And it's not like you would write `new Text({ text: 'blah' })` that much, is it?

juliussohn:
When will this fix be released?

juliussohn:
I am still getting in `ObjC method new requires 0 arguments, but JavaScript passed 1 argument` when trying to access `textLayer.fragments`

mathieudutour:
what sketch version are you using?

juliussohn:
52.6

PR #281, opened 2018-11-08
Add verticalAlignment to Text

andrewstart:
Adds `verticalAlignment` and `VerticalAlignment` to `Text`.

I am reading the vertical alignment of text fields in a plugin, and figured that I would make this as it is a small feature that was missing.
Note: As of Sketch 52.3, `MSImmutableTextLayer` appears to not implement a getter for `verticalAlignment`, so as currently written this feature isn't safe. I could add a check to make sure that the property exists, or the PR can wait for an implementation to be added to Sketch.

### DIFF
--- a/Source/dom/layers/Text.js
+++ b/Source/dom/layers/Text.js
@@ -44,7 +44,7 @@ const VerticalTextAlignment = {
 }
 
 export const VerticalTextAlignmentMap = {
-  left: 0, // Visually left aligned
+  top: 0, // Visually left aligned
   center: 1, // Visually centered
   bottom: 2, // Visually bottom aligned
 }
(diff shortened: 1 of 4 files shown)


Source/dom/layers/Text.js
```diff
@@ -36,6 +36,19 @@ export const TextAlignmentMap = {
...
 
+// Mapping between vertical text alignment names and values.
+const VerticalTextAlignment = {
+  top: 'top', // Visually top aligned
+  center: 'center', // Visually center aligned
+  bottom: 'bottom', // Visually bottom aligned
+}
+
+export const VerticalTextAlignmentMap = {
+  left: 0, // Visually left aligned
```
bomberstudios: Is this `left` correct? Shouldn't this be `top`?

andrewstart: Whoops, missed that when adjusting after copying and pasting.

PR #302, opened 2018-11-29
add session variables methods

### LINKED ISSUE #301: add method to set and get a session variable
Right now there is no easy way to store a variable for the time Sketch is running.

Let's fix that.

mathieudutour:
fix #301


bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

docs/api/Settings.md
```diff
@@ -129,3 +129,32 @@ Store a value of a setting for a given key on a specific document.
...
 | key<span class="arg-type">string - required</span>                     | The setting to set.                       |
 | value<span class="arg-type">any - required</span>                      | The value to set it to.                   |
+
+## Get a session variable
+
+```js
+var myVar = Settings.sessionVariable('myVar')
+```
+
+Return the value of a variable which is _not_ persisted when Sketch closes but is persisted when the plugin finishes to run. It is useful when you want to keep a value between plugin's runs.
```
robintindale: this seems like a confusing way to word it? Saying what it's *not* before saying what it *is*

mathieudutour: Good point!

Source/settings/Settings.js
```diff
@@ -3,9 +3,8 @@ import { isWrappedObject, getDocumentData } from '../dom/utils'
 
 function getPluginIdentifier() {
   if (!__command.pluginBundle()) {
-    throw new Error(
-      'It seems that the command is not running in a plugin. Bundle your command in a plugin to use the Settings API.'
-    )
+    // if we run a script from the Run Script panel, it won't have a bundle
+    return 'com.sketchapp.temporary-script'
```
bomberstudios: I wonder if this change has any impact anywhere else? I can't think of any scenario in which something would be expecting an error, so this is probably safe, but I thought I'd ask

Source/settings/Settings.js
```diff
@@ -141,3 +140,32 @@ export function setDocumentSettingForKey(document, key, value) {
...
+  }
+  return JSON.parse(value)
+}
+
+export function setSessionVariable(key, value) {
+  const threadDic = NSThread.mainThread().threadDictionary()
+
+  const stringifiedValue = JSON.stringify(value, (k, v) => util.toJSObject(v))
+  if (!stringifiedValue) {
+    threadDic.removeObjectForKey(
```
bomberstudios: Maybe I'm reading this wrong, but… if a plugin tries to store something in the session variable that can't be stringified, we erase the existing contents of that variable? Isn't this a dangerous thing to do? If I already have something there and there's a bug in my code that tries to store something that can't be stringified, I'll lose the existing value… 🤔

Source/settings/__tests__/Settings.test.js
```diff
@@ -1,5 +1,4 @@
 /* globals expect, test */
-import { isRunningOnJenkins } from '../../test-utils'
```
bomberstudios: 👏 for less code 🙂

Source/settings/Settings.js
mathieudutour: Yeah that's only happening in the Run Script panel so I'd say it's safe to assume nobody care

Source/settings/Settings.js
mathieudutour: if it can't be stringified, it will crash before trying to set anything so I don't think it's an issue.

But actually most of the times _I_ use the thread dictionary, it's to store a native object that I don't want to stringify 🤔

Issue #292, opened 2018-11-27
Add JSON archiving/unarching


WCByrne:
Any plans to add archiving to the JS API to replace direct access to `MSJSONDataArchiver`? The current toJSON doesn't seem to get everything to recreate the object, example [here](https://sketchplugins.com/d/1127-tojson-couldn-t-get-the-complete-info). 

Maybe this doesn't make sense but for my uses I've wrapped the object in a dict to include the version meta with the object. 
```javascript
{
    "meta": {
        "version": 112
        ...
    },
    object: {
        ...
    }
}
```

mathieudutour:
What's the use case for using `MSJSONDataArchiver`?

mathieudutour:
The thing is that the JSON produced by the archiver might change with every sketch version and the goal of this API is to be stable

mathieudutour:
Yeah, happy to go with something like that if we make it clear that the output is likely to change in the docs.

How about `getSketchJSON` to match `.sketchObject`? I wouldn't mind returning a JSON object, or is there any use case for returning just the string? It seems that everybody will just parse it anyway

mathieudutour:
But tbh, I'm still wondering about the use cases for it 🤔 
The only one I have in mind is having a algorithm that can run on both the JSON file and in a plugin.
My secret plan is to create a nodejs lib that allow to manipulate the JSON file with the same API as this one but that's more long term so it might be worth adding this for now

robintindale:
I guess one use-case is allowing people to access things that aren't in the API yet without need knowledge of the internals - they can get the `sketchJSON` and find what they need. This could be a good way to steer people away from `sketchObject` and needing the headers

Is it actually possible for people to `setSketchJSON`? Obviously could get weird if they give malformed JSON, but would it work to use `JSONUnarchiver` then replace the properties of the existing layer with the new one if the unarchive is successful?

mathieudutour:
it would be possible yes (I think it's `MSJSONDictionaryUnarchiver` that we'd need) but you need to pass a document version to it and that's a pretty big can of worms that I don't really want to open

WCByrne:
My use case is probably not recommended but I'm storing the archive outside the document then unarchiving it to use again later (often in another document). Hence keeping track of the version with each archive. 

Archive doesn't seem like the right term in JS but does it help to think about it more like that? That is, a JSON representation that can be used to recreate this object later.

WCByrne:
[...]

Issue #185, opened 2018-04-27
return a Buffer from an export instead of writing to file


jay7n:
Hi guys, 

Is there a way (such an API) to acquire the thumbnail view's data from the layer inspector, which is highlighted in the below picture with a red rectangle?   I need this kind of data ( as a NSImage type) to do some image comparisons. 

![default](https://user-images.githubusercontent.com/1228472/39345492-46fb46f0-4a1b-11e8-8f5b-6e6aea73dffd.png)

Thanks in advance !

mathieudutour:
the [`export`](https://developer.sketchapp.com/reference/api/#export) API only write to a file. We discussed with @bomberstudios about a way to return an `NSData` if an option is specified but it's not possible with the API yet

jay7n:
Is this (returning NSData from the export) being developed ? would appreciate it if I could know how long I have to wait to use it ... honestly I got stuck here and can't step any further. 

If it's impossible to expect it in a near future, do you have any kind suggestions for some alternative ways to do the job ? 

Thanks for your time again ~

robintindale:
You can use the old API for now, being aware of the disadvantages
```
  var ancestry = MSImmutableLayerAncestry.ancestryWithMSLayer_(layer);
  var exportRequest = MSExportRequest.exportRequestsFromLayerAncestry_(ancestry).firstObject();
  exportRequest.format = 'png';
  var rend1 = MSExportManager.alloc().init().rendererForRequest(exportRequest);
  var rep1 = rend1.bitmapImageRep();
  var data = rep1.representationUsingType_properties_(NSPNGFileType, nil);
```

jay7n:
thank you @robintindale I'll give it a try

PR #306, opened 2018-12-05
JSON Object Archives

WCByrne:
This is a basic implementation of #292 and allows:
* Fetching a dictionary archive of an object
* Restoring an object from an archive dictionary

Not sure if this is the best way to archive/unarchive objects but it has been working for me in my plugin. Also very open to naming suggestions.


bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/WrappedObject.js` changed, but not its tests. That's OK as long as you're refactoring.
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/wrapNativeObject.js` changed, but not its tests. That's OK as long as you're refactoring.
  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

Source/dom/WrappedObject.js
```diff
@@ -102,6 +102,26 @@ export class WrappedObject {
...
+  jsonArchive() {
+    const archiver = MSJSONDataArchiver.new()
+    archiver.archiveObjectIDs = false
+    const aPtr = MOPointer.alloc().init()
+    const obj = this.isImmutable()
+      ? this.sketchObject
+      : this.sketchObject.immutableModelObject()
+    archiver.archivedDataWithRootObject_error(obj, aPtr)
+    if (aPtr.value()) {
+      console.error(`Archive error ${aPtr.value()}`)
```
mathieudutour: Extra console.error, we don’t want to log when we throw

Source/dom/__tests__/Archive.test.js
```diff
@@ -0,0 +1,17 @@
+/* globals expect, test */
+
+import { objectFromArchive } from '../wrapNativeObject'
+
+test('should archive object', () => {
+  const object = MSLayer.new()
```
mathieudutour: Probably want to create a Layer here: `const object = new Shape()` (and you need to import `Shape` obviously)

Source/dom/wrapNativeObject.js
```diff
@@ -52,3 +52,17 @@ export function wrapObject(object, defaultType) {
...
   return Factory.create(type || defaultType, rest)
 }
+
+export function objectFromArchive(archive, version) {
+  const v = version || MSArchiveHeader.metadataForNewHeader().version
+  let object = MSJSONDictionaryUnarchiver.unarchiveObjectFromDictionary_asVersion_corruptionDetected_error(
+    archive,
+    v,
+    null,
+    null
```
mathieudutour: Can you add some error handling?

Source/dom/WrappedObject.js
```diff
@@ -102,6 +102,26 @@ export class WrappedObject {
     return /Immutable/.test(String(this.sketchObject.className()))
   }
 
+  jsonArchive() {
```
mathieudutour: I’d rather go with `getSketchJSON`. “Archive” is not really explicit if you don’t know the underlying native API.

Source/dom/index.js
```diff
@@ -23,7 +23,10 @@ export { HotSpot } from './layers/HotSpot'
 export { Slice } from './layers/Slice'
 
 export { Types } from './enums'
-export { wrapObject as fromNative } from './wrapNativeObject'
+export {
+  wrapObject as fromNative,
+  objectFromArchive as fromArchive,
```
mathieudutour: Same, `fromSketchJSON`

WCByrne:
I like the idea of considering this an export 💯 . I was thinking there was a better place for this to live but went with the easier option to start.

WCByrne:
@mathieudutour Since the `exportObject` saves exported data to a path, do you think it makes sense to do incorporate this export into the same method? Especially considering the output is a list of desired types.

Thinking we could...
1. If the output === `json` or contains `json`, return that immediately (i.e. you can't have multiple types with json)
1. Make a new export method next to `exportObject` 
1. Save the JSON - I don't think this is a good option especially considering the other formats happen inside the native exporter.

Option 1 seems like the cleanest to me.

mathieudutour:
Ah yes I should have explained a bit further. There is this other issue that I want to work on as well: 
https://github.com/BohemianCoding/SketchAPI/issues/185.

I was thinking of the following:
- if `output` is flash, then return the data instead of writing it to disk.
- if there is an output, then write the json string to `${output}/export.json`

I realised that there will be a disparity in the API until #185 is done but we are building for the future :)

WCByrne:
Sorry I meant `format` equals or contains `json`. 

Is there any concern in changing the behavior of a falsey `output`? Right now it defaults to the `~/Documents/Sketch Exports` right? 

As for implementing, is there a way to get the data back from the exporter being used or does that need to be handled differently? I use `MSExportRequest` in my plugin... 🤔 

Also, thoughts on returning in the case of `!option` and the fact that formats can be an array? Would that be considered invalid? Return the first one?

mathieudutour:
- If `format` is `json`, then only export use the new code path
- if `format` contain `json`, then we remove it from the `format` array, pass it to the old code path and use the new code path as well

- if `output` is falsy (it will default to a path but can still be overwritten by a falsy value), then we return the value (only for the json for now, the implementation for the images will be in another PR)
- if it’s a path, then we write to a file

WCByrne:
Sorry for the broken/WIP, needed to move to a different computer. Will finish up shortly.

Source/dom/export.js
```diff
@@ -46,58 +46,142 @@ export const DEFAULT_EXPORT_OPTIONS = {
...
+ * @returns If an output path is not set, the data is returned
  */
 export function exportObject(object, options) {
-  const merged = { ...DEFAULT_EXPORT_OPTIONS, ...options }
+  if (!object) {
+    throw new Error('No object provided to export')
+  }
+  let { formats } = options
+  if (typeof formats === 'string') {
+    formats = [formats]
```
mathieudutour: So I think this should be `formats = formats.split(‘,’).map(format => format.trim())` because it is supposed to be a string of comma separated values. I’m fine with accepting arrays tho, it’s much cleaner. 

But the docs need to be changed and we still need to pass a string to the native method.

WCByrne: Got it. I wasn't actually 100% sure if it was supposed to be an array or string. I think the docs say a "list".

Source/dom/export.js
```diff
@@ -46,58 +46,146 @@ export const DEFAULT_EXPORT_OPTIONS = {
...
+ *
+ * @returns If an output path is not set, the data is returned
  */
 export function exportObject(object, options) {
-  const merged = { ...DEFAULT_EXPORT_OPTIONS, ...options }
+  if (!object) {
+    throw new Error('No object provided to export')
+  }
+  let exportJSON = false
+  let { formats } = options
```
mathieudutour: I believe you need `options || {}` otherwise it will throw when there is no options

```suggestion
  let { formats } = options || {}
```

Source/dom/export.js
```diff
@@ -46,58 +46,146 @@ export const DEFAULT_EXPORT_OPTIONS = {
...
+  if (formats) {
+    const idx = formats.indexOf('json')
+    if (idx >= 0) {
+      formats.splice(idx, 1)
+      exportJSON = true
+    }
+    formats = formats.join(',')
+  }
+
+  const merged = {
```
mathieudutour: Can we call it optionsWithDefaults or something more explicit?

Source/dom/export.js
```diff
@@ -46,58 +46,146 @@ export const DEFAULT_EXPORT_OPTIONS = {
...
+    ...DEFAULT_EXPORT_OPTIONS,
+    ...options,
+    formats,
+  }
+  const objectsToExport = (Array.isArray(object) ? object : [object]).map(
+    o => (isWrappedObject(o) ? o.sketchObject : o)
+  )
+  if (!objectsToExport) {
```
mathieudutour: It will always be truthy here

```suggestion
  if (!objectsToExport.length) {
```

Source/dom/export.js
```diff
@@ -46,58 +46,146 @@ export const DEFAULT_EXPORT_OPTIONS = {
...
+  function archiveNativeObject(nativeObject) {
+    const archiver = MSJSONDataArchiver.new()
+    archiver.archiveObjectIDs = true
+    const aPtr = MOPointer.alloc().init()
+    const obj = nativeObject.immutableModelObject
+      ? nativeObject.immutableModelObject()
+      : nativeObject
+    archiver.archivedDataWithRootObject_error(obj, aPtr)
+    if (aPtr.value()) {
+      throw Error('Archive error')
```
mathieudutour:
```suggestion
      throw Error(`Couldn’t create the JSON string: ${aPtr.value()}`)
```

Source/dom/export.js
```diff
@@ -46,58 +46,146 @@ export const DEFAULT_EXPORT_OPTIONS = {
...
+    }
+    return archiver.archivedData()
+  }
+
+  // Return data if no output directory specified
+  if (!merged.output) {
+    const firstObject = objectsToExport[0]
```
mathieudutour: Why only the first object? Let’s export all of them and return an array

Source/dom/export.js
```diff
@@ -46,58 +46,146 @@ export const DEFAULT_EXPORT_OPTIONS = {
...
+      const writeOptions = merged.overwriting
+        ? 0
+        : NSDataWritingWithoutOverwriting
+      const ptr = MOPointer.new()
+      if (!data.writeToURL_options_error(url, writeOptions, ptr)) {
+        throw new Error(`Error writing json file ${ptr.value()}`)
+      }
+    })
+    // If only JSON stop to bypass the exporters PNG default
+    if (merged.formats.length == 0) {
```
mathieudutour: formats can be undefined

```suggestion
    if (!(merged.formats || {}).length) {
```

Source/dom/export.js
WCByrne: Sure thing. I just stuck with the previous name.

Source/dom/export.js
WCByrne: That felt a little weird in the case of the image data but I agree it is the less confusing option. I'll fix it

WCByrne: Looking to the future, any thoughts on multiple formats? My instinct is to limit it to a single type if `!output`. The only good response I can think of otherwise is an array of objects like:
```
[ 
    { 'jpg' : <data>, png: <data> },
    { 'jpg' : <data>, png: <data> }
]
```

mathieudutour: That’s a good question, I was actually thinking about returning an NSImage so it wouldn’t care about the format.

So let’s do this: if there are more than one format and output is falsy, then we throw an error. We only authorise 1 format if there is no output

WCByrne:
Am I missing something about running tests? with `npm run test` I get `Xcode build not found`.  With `npm run test:no-variant` I get `Error: couldn't find the test results`.

I can see the tests running in Sketch and saving things to my Desktop 🤔

mathieudutour:
Hum not sure. You can `non run test:no-variant` IIRC

WCByrne:
That's what gives `Error: couldn't find the test results`...

mathieudutour:
(I'm not forgetting this, I'll set time aside to test it properly next week)

mathieudutour:
I opened a PR on your repo to refactor a bit: https://github.com/WCByrne/SketchAPI/pull/2/commits/b6c33438eb2b42d6eeaa61395881f8b3034d31c8

Feel free to review it and merge it and then we can merge this one

WCByrne:
Thanks for cleaning that all up. I removed what like there was some leftover exporting happening at the end of the main export function. This PR should have the latest now.

WCByrne:
Heads up. Assuming the betas pickup the local dev API from defaults, this doesn't seem to be working in 53 😕

PR #317, opened 2018-12-20
return a mocked up library if there is a foreign object

mathieudutour:


bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/layers/SymbolMaster.js` changed, but not:
  - 📚 its docs
- 🧪 its tests

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/models/SharedStyle.js` changed, but not:
  - 📚 its docs
- 🧪 its tests

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

mathieudutour:
> What will happen if the user tries to do something with the returned library?

It won't do anything. That's the same with pretty much any invalid library tho

> Would it be safer to have a method like `isLocal` or `isForeign` for this?

I'd rather not introduce more terms if they are not needed

PR #325, opened 2019-01-05
fix wether typos and #shared-style links

robintindale:
* fixed wether typos, should be Whether
* fix #sharedstyle links to be the correct #shared-style


mathieudutour:
Thanks! The `gh-pages` is generated from the `develop` branch, could you rebase your changes there?

PR #330, opened 2019-01-07
Document LayersResized

mikeabdullah:
Connect to BohemianCoding/Sketch#21556.


mikeabdullah:
Also, does it mean anything that the above tests have failed?

mathieudutour:
no it's just the preview that failed, not sure why but it's not important

Issue #44, opened 2016-11-21
Add support for document colors


kgn:
Both colors and gradients

Proposal:

Application.globalColors - get/set
Application.globalGradients - get/set

kgn:
I think with this it would be good to have a wrapper class for `Color` that's used with these.

WCByrne:
Any plans for this one? I can take a stab at it

I was curious if there were any plans to add IDs to document colors to make it possible to reference them but considering the JS API is only representing colors as hex values I'm guessing that is a no. @mathieudutour would you expect access to the document colors to be the same or something more robust/flexible?

mathieudutour:
We do have some plans regarding document's and global colors, that's why I was waiting a bit it before adding something to the API.

> I was curious if there were any plans to add IDs to document colors to make it possible to reference them
That would be a change in Sketch so we can't just add it in the API.

So I was thinking about having:
```
sketch.globalColors = [
 {
    color: Color,
    // we might add some stuff there later
  }
]
sketch.globalGradients = [
  {
    gradient: Gradient,
    // we might add some stuff there later
  }
]

document.colors = [...]
document.gradients = [...]
```

What do you think?

WCByrne:
The API makes sense to me.  I assume `Color` is a color object? If so, could the list just be `[Color]`/`Gradient`? It's not clear what "some stuff" would be, extra color information or something specific to the color in the list?

Let me know if I can help.

mathieudutour:
`Color` would indeed be that object: https://github.com/BohemianCoding/SketchAPI/blob/develop/Source/dom/style/Color.js and `Gradient` that one: https://github.com/BohemianCoding/SketchAPI/blob/develop/Source/dom/style/Gradient.js.

The extra information might be an id or a name for example. Once again, that's not something that is currently implemented in Sketch so we just have to be careful here not to introduce breaking changes in the future.

WCByrne:
👍 Got it. I didn't see that class there.

Probably makes the most sense to figure out how the internals will work before getting too far into it. i.e will the document colors be a sort of SharedColor object like SharedStyles are? Any timeline for that yet?

WCByrne:
In light of a breaking change with the internal API here it would be nice to push this forward. 

Regarding the extra information, is there a reason that info would need to go in an object around the color rather than directly on the color object itself?

mathieudutour:
So actually let’s do it like this:
```js 
sketch.globalColors = [
 {
    name: string,
[...]

PR #345, opened 2019-01-17
Colors

WCByrne:
Issue: https://github.com/BohemianCoding/SketchAPI/issues/44

Adds API access to document and global color/gradient assets using new ColorAsset and GradientAsset classes. 

### Global
This is only read  at the moment
```javascript
sketch.globalColors() // [ColorAsset]
sketch.globalGradients() // [GradientAsset]
```

### Document
```javascript
let colors = doc.colors // [ColorAsset]
let gradients = doc.gradients // [GradientAsset]

colors[0].color // hex
gradients[0].gradient // Gradient

 // Push a color as a hex string or native object (i.e NSColor, MSColor, MSColorAsset, etc)
// Name defaults to null
colors.push('#AABBCCFF')

// Push a gradient as dictionary or existing Gradient object. 
gradients.push({
  gradientType: 'Linear',
  ...
})


// Optionally include a name
colors.push({
  name: 'My Color',
  color: '#112233', // Or native
})
gradients.push({
  name: 'My Gradient',
  gradient: {}
})
```

### TODO
* Should you be able to set global colors/gradients?


Source/dom/models/Document.js
```diff
@@ -470,3 +471,46 @@ Document.define('path', {
...
+  array: true,
+  get() {
+    if (!this._object) {
+      return []
+    }
+    const documentData = this._getMSDocumentData()
+    const colorAssets = toArray(documentData.assets().colorAssets())
+    return colorAssets.map(a => ({
+      color: Style.colorToString(a.color()),
+      name: a.name(),
```
mathieudutour: So thinking about it, I'm not sure it's enough to return a simple object here:
```
doc.colors[0].name = 'new name'
```
is not going to work for example.

So I think we need to create a new wrapped object.

But otherwise, all those methods look good 👍

WCByrne: Ah, yup. 

I put a ColorAsset model in but then remembered gradients. Any thoughts on dealing with those 2? 

I'm Thinking 2 distinct classes would be the least confusing. Another potential option could be to abstract the internals with one class that wraps either "asset" type and exposes `gradient` and `color` properties where one always returns null/undefined - maybe with an "assetType".

mathieudutour: That’s the 2 options I had in mind as well, not sure what’s the best. 

I was also thinking about having `value` instead of `color` so it can be the same for all assets. 

But maybe having 2 separate classes will be more flexible for the future

WCByrne:
Ready for another review here. I considers breaking the new classes out into separate files but it seemed like extra clutter - happy to do so if that’s better.

Source/dom/models/Asset.js
```diff
@@ -0,0 +1,126 @@
+import { isNativeObject } from 'util'
+import { WrappedObject, DefinedPropertiesKey } from '../WrappedObject'
+import { Types } from '../enums'
+import { Factory } from '../Factory'
+import { Color, colorToString } from '../style/Color'
+import { Gradient } from '../style/Gradient'
+import { isWrappedObject } from '../utils'
+
+export class Asset extends WrappedObject {}
+
```
mathieudutour: let's do `delete Asset[DefinedPropertiesKey].id` here instead of below
```suggestion
Asset[DefinedPropertiesKey] = { ...WrappedObject[DefinedPropertiesKey] }
delete Asset[DefinedPropertiesKey].id

```

Source/dom/models/Asset.js
```diff
@@ -0,0 +1,126 @@
...
+      const c = Color.from(object)._object.newMutableCounterpart()
+      nativeAsset = MSColorAsset.alloc().initWithAsset_name(c, null)
+    }
+
+    return ColorAsset.fromNative(nativeAsset)
+  }
+}
+
+ColorAsset.type = Types.ColorAsset
+ColorAsset[DefinedPropertiesKey] = { ...WrappedObject[DefinedPropertiesKey] }
```
mathieudutour:
```suggestion
ColorAsset[DefinedPropertiesKey] = { ...Asset[DefinedPropertiesKey] }
```

Source/dom/models/Asset.js
```diff
@@ -0,0 +1,126 @@
...
+Factory.registerClass(ColorAsset, MSColorAsset)
+delete ColorAsset[DefinedPropertiesKey].id
```
mathieudutour:
```suggestion
```

Source/dom/models/Asset.js
```diff
@@ -0,0 +1,126 @@
...
+    const nativeAsset = MSGradientAsset.alloc().initWithAsset_name(
+      gradient._object,
+      name || null
+    )
+    return GradientAsset.fromNative(nativeAsset)
+  }
+}
+
+GradientAsset.type = Types.GradientAsset
+GradientAsset[DefinedPropertiesKey] = { ...WrappedObject[DefinedPropertiesKey] }
```
mathieudutour:
```suggestion
GradientAsset[DefinedPropertiesKey] = { ...Asset[DefinedPropertiesKey] }
```

Source/dom/models/Asset.js
```diff
@@ -0,0 +1,126 @@
...
+Factory.registerClass(GradientAsset, MSGradientAsset)
+delete GradientAsset[DefinedPropertiesKey].id
```
mathieudutour:
```suggestion
```

docs/api/Document.md
```diff
@@ -220,6 +220,30 @@ A method to help find a symbol master in the document.
 
 Return a [SymbolMaster](#symbolmaster) object or `undefined` if it's not found.
 
+## Get all the Document Colors
+
+```javascript
+var colors = document.colors
+```
+
+A method to get all color assets defined in the document.
```
mathieudutour: That's not a method, it's a property so it should go in the properties' table

Source/dom/models/Asset.js
```diff
@@ -0,0 +1,126 @@
+import { isNativeObject } from 'util'
```
mathieudutour: maybe instead of having one big file with 3 definitions, we can create a folder `Asset` with a file `index.js` exporting the `ColorAsset` and `GradientAsset` defined in their own file

WCByrne: Are you thinking that asset level would be in `/models` or on the same level as `/style`?

mathieudutour: I'd put it on the same level as `/style` but I really don't mind either

WCByrne: Went ahead and moved everything to `/assets` similar to `/layers`.

Source/dom/index.js
```diff
@@ -10,6 +10,7 @@ export { Document, getDocuments, getSelectedDocument } from './models/Document'
 export { Library, getLibraries } from './models/Library'
 export { SharedStyle } from './models/SharedStyle'
 export { Rectangle } from './models/Rectangle'
+export { ColorAsset, GradientAsset } from './assets'
```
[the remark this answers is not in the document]
WCByrne: I had the same thought, I guess not.

WCByrne:
Let me know if there is anything else to fix/improve here

docs/api/Document.md
```diff
@@ -15,6 +15,8 @@ A Sketch document.
 | id<span class="arg-type">string</span>             | The unique ID of the document. |
 | pages<span class="arg-type">[Page](#page)[]</span> | The pages of the document.     |
 | path<span class="arg-type">string</span>           | The path to the document (or the appcast URL in case of a Document from a remote Library.     |
+| colors<span class="arg-type">[GraientAsset]</span>           | A list of color assets defined in the document. Mutating the returned array will update the document colors.     |
```
mathieudutour: The documentation for the asset is missing

Source/dom/globalAssets.js
```diff
@@ -0,0 +1,22 @@
+import { toArray } from 'util'
+import { ColorAsset, GradientAsset } from './assets'
+
+/**
+ * Access any global colors
+ *
+ * @return {Array<ColorAsset>} A list of color assets defined globally
+ */
+export function globalColors() {
```
mathieudutour: can we call it `getGlobalColors`?

mathieudutour: The documentation for this is missing

docs/api/Document.md
WCByrne: I wasn't sure where it should go

Source/dom/globalAssets.js
WCByrne: Again, I wasn't sure where it should go

docs/api/Document.md
WCByrne: Same with the global getters

mathieudutour: So I'd create a new file in the `Models` section called `Asset`. Then each asset is a `##` header and the global getter a `###` under it.

What do you think?

WCByrne: Seems like it makes sense. I haven't worked with the docs enough yet to picture what that would actually look like. I can try it. 

It's not totally clear in the Readme if I can build and view the docs locally. Is that possible?

WCByrne: Ok, added an Assets section to the docs.

PR #350, opened 2019-01-22
fix the order of overrides

### LINKED ISSUE #343: Return symbolInstance.overrides in order as displayed in inspector panel
I'd expect `var overrides = symbolInstance.overrides` to give an array of overrides **in the same order as they are displayed in the inspector panel**.

When a symbol does not contains any nested foreign symbols, then `symbolInstance.overrides` appears to return an array with each override listed in reverse order. But if the symbol contains nested symbols from a library then it looks like the library overrides are listed first so I can't see an obvious way of sorting the overrides into the display order.

Not sure if this is a bug or a feature request.

mathieudutour:
fix #343

### DIFF
--- a/Source/dom/layers/SymbolInstance.js
+++ b/Source/dom/layers/SymbolInstance.js
@@ -120,17 +120,11 @@ SymbolInstance.define('overrides', {
     if (!this._object.availableOverrides) {
       return undefined
     }
-    const overrides = toArray(this._object.availableOverrides())
-
-    // recursively find the overrides
-    function findChildrenOverrides(instance) {
-      const children = toArray(instance.children())
-      children.forEach(c => {
-        overrides.push(c)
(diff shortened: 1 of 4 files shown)


bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/layers/SymbolInstance.js` changed, but not:
  - 📚 its docs
- 🧪 its tests

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/layers/SymbolMaster.js` changed, but not:
  - 📚 its docs

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #361, opened 2019-01-31
export returns a buffer if output is false

### LINKED ISSUE #185: return a Buffer from an export instead of writing to file

mathieudutour:
fix #185


Source/dom/__tests__/export.test.js
```diff
@@ -65,6 +62,7 @@ test('should fail with to return with mulitple formats', () => {
       formats: ['png', 'json'],
       output: false,
     })
+    expect(false).toBe(true)
```
bomberstudios: what? 😅

Source/dom/__tests__/export.test.js
```diff
@@ -44,18 +44,15 @@ test('Should fail with no object provided', () => {
...
+    parent: document.selectedPage,
+  })
+  const buffer = exportObject(object, {
+    formats: 'png',
+    output: false,
+  })
+  expect(Buffer.isBuffer(buffer)).toBe(true)
 })
 
 test('should fail with to return with mulitple formats', () => {
```
bomberstudios: Typo
```suggestion
test('should fail with to return with multiple formats', () => {
```

Issue #366, opened 2019-02-06
Add ability to import files


WCByrne:
When pasting vector graphic files into Sketch they are converted to native layers. It would be great if to be able to tap into this programmatically via the API.

Perhaps this could be in the form of a method to import any importable file. It could handle pdfs, svgs, bitmap and returns the layers. Something along the lines of:
```javascript
function layersFromFileData(data) -> Layer
```

WCByrne:
If anyone can shed some light on the internals for accomplishing this I would be happy to work on the API implementation. 

The best I've come across so far is from this [forum post](https://sketchplugins.com/d/267-import-svg/14) which has this code snippet:
```javascript
var svgString = NSString.stringWithString('<svg width="200px" height="100px" viewBox="0 0 200 100" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><rect fill="#000000" x="0" y="0" width="200" height="100"></rect></svg>');
var svgData = svgString.dataUsingEncoding(NSUTF8StringEncoding);
var svgImporter = MSSVGImporter.svgImporter();
svgImporter.prepareToImportFromData(svgData);
var svgLayer = svgImporter.importAsLayer();
svgLayer.setName("LayerName");
context.document.currentPage().addLayers([svgLayer]);
```

I haven't come across anything for PDFs yet.

mathieudutour:
instead of `MSSVGImporter.svgImporter()`, you have `MSPDFImporter.pdfImporter()` or `MSPDFImporter.epsImporter()`. The rest of the API is the same

PR #369, opened 2019-02-08
File import

WCByrne:
ℹ️ Address #366  
⚠️ Requires #372

### DIFF
--- a/Source/dom/__tests__/import.test.js
+++ b/Source/dom/__tests__/import.test.js
@@ -1,0 +1,53 @@
+/* globals expect, test */
+import { createLayerFromData } from '../import'
+import { Rectangle } from '../models/Rectangle'
+import { exportObject } from '../export'
+import { Shape } from '../layers/Shape'
+
+test('should create svg layer', () => {
+  const svgString =
+    '<svg width="200px" height="100px" viewBox="0 0 200 100" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><rect fill="#000000" x="0" y="0" width="200" height="100"></rect></svg>'
+
+  const group = createLayerFromData(svgString, 'svg')
+  expect(group.type).toEqual('Group')
+  expect(group.layers.length).toEqual(1)
+  expect(group.layers[0].type).toEqual('ShapePath')
+  expect(group.layers[0].shapeType).toEqual('Rectangle')
+  expect(group.layers[0].frame.width).toEqual(200)
+  expect(group.layers[0].frame.height).toEqual(100)
+})
+
+test('should create page from PDF', (context, document) => {
+  const layer = new Shape({
+    parent: document.selectedPage,
+    frame: new Rectangle(0, 0, 200, 100),
+    style: {
+      fills: [
+        {
+          color: '#888888',
+        },
+      ],
+    },
+  })
+  const buffer = exportObject(layer, {
+    formats: 'pdf',
+    output: null,
+  })
+  const group = createLayerFromData(buffer, 'pdf')
+  expect(group.type).toEqual('Group')
+  expect(group.frame.width).toEqual(200)
+  expect(group.frame.height).toEqual(100)
+})
+
+test('should create Image layer from image', (context, document) => {
+  const layer = new Shape({
+    parent: document.selectedPage,
+    frame: new Rectangle(0, 0, 200, 100),
+  })
+  const buffer = exportObject(layer, {
+    formats: 'png',
+    output: null,
+  })
+  const page = createLayerFromData(buffer, 'image')
+  expect(page.type).toEqual('Image')
+})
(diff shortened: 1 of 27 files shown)


bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/import.js` changed, but not:
  - 📚 its docs

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

Source/dom/import.js
```diff
@@ -0,0 +1,35 @@
+import { wrapNativeObject } from './wrapNativeObject'
+
+function _importWithImporter(data, importer) {
+  importer.prepareToImportFromData(data)
```
mathieudutour: I'd rather consider `data` to be a Buffer (and also handle the case where it's an NSData) because `@skpm/fs` returns a Buffer (available through `require('buffer').Buffer`: https://github.com/skpm/buffer)

Source/dom/import.js
```diff
@@ -0,0 +1,35 @@
...
+    case 'pdf':
+      return _importPDF(data)
+    case 'eps':
+      return _importEPS(data)
+    default:
+      throw new Error(`Can't import ${type}`)
+  }
+}
+
+export const layerFromString = (str, type) => {
```
mathieudutour: I don't think we need this tbh

Source/dom/import.js
```diff
@@ -0,0 +1,35 @@
...
+  return _importWithImporter(data, MSSVGImporter.svgImporter())
+}
+function _importPDF(data) {
+  return _importWithImporter(data, MSPDFImporter.pdfImporter())
+}
+function _importEPS(data) {
+  return _importWithImporter(data, MSPDFImporter.epsImporter())
+}
+
+export const layerFromData = (data, type) => {
```
mathieudutour: let's call it `createLayerFromData` to make it clear that it's creating a layer

Source/dom/import.js
```diff
@@ -0,0 +1,35 @@
...
+  switch (type) {
+    case 'svg':
+      return _importSVG(data)
+    case 'pdf':
+      return _importPDF(data)
+    case 'eps':
+      return _importEPS(data)
+    default:
```
mathieudutour: it'd be super easy to handle bitmaps as well: `return Image({image: data})`

Source/dom/import.js
WCByrne: Removed. This was just to make it easy to create from SVG strings but it seems that can happen with a Buffer now.

Source/dom/__tests__/import.test.js
```diff
@@ -0,0 +1,36 @@
...
+test('should create layer from PDF', (context, document) => {
+  const layer = new Shape({
+    parent: document.selectedPage,
+    frame: new Rectangle(0, 0, 200, 100),
+  })
+  const buffer = exportObject(layer, {
+    format: 'pdf',
+    output: null,
+  })
+  const group = createLayerFromData(buffer, 'pdf')
```
WCByrne: For some reason `pdf` always returns a page. `createLayerFromData(svgString, 'pdf')` It also works somehow, again, returning a `Page`. Not sure what is going on here.

mathieudutour: the pdf importer creates an artboard per pdf page so it makes sense that it returns a page.

not sure about the `createLayerFromData(svgString, 'pdf')` tho, does it actually parses the svg?

WCByrne: Ah, that makes sense. Should that Page be returned here? It is technically a layer but that may be a little confusing. 

The `svgString` case is just creating an empty page, I guess I thought it would fail. Maybe this function should fail if there are no layers on the page 🤔

WCByrne: It looks like passing pdfData with `pdf` as the import type isn't working either. The returned page is empty.

Source/dom/import.js
WCByrne: Shouldn't Image accept a Buffer for its data? That doesn't appear to be the case right now.

mathieudutour: Ah I thought I did it but I might not have pushed it

WCByrne: If not, I've added it in to try it out.

mathieudutour: found my branch: https://github.com/BohemianCoding/SketchAPI/pull/372

WCByrne:
Ok, everything seems to be working now with a few notes:
* Needs #372 
* The [PDF import test](https://github.com/BohemianCoding/SketchAPI/pull/369/files#diff-e6137b3fe9213bd136e7494eb0e3445fR24) only seems to work if the exported layer has a style. Not sure if that is a bug in the importer or not that may be symptomatic of an issue that may affect other PDFs.

PR #373, opened 2019-02-11
fix setting flow as undefined

mathieudutour:


bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  `Source/dom/layers/Layer.js` changed, but not:
  - 📚 its docs

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

Source/dom/layers/Layer.js
```diff
@@ -252,13 +252,13 @@ Layer.define('selected', {
...
   get() {
-    return wrapObject(this._object.flow())
+    return Flow.fromNative(this._object.flow())
   },
   set(_flow) {
     if (this.isImmutable()) {
       return
     }
-    const flow = Flow.from(_flow)
+    const flow = _flow ? Flow.from(_flow) : null
```
christianklotz: Wouldn't `flow.sketchObject` in the following line throw in error if `flow` turns out to be `null`?

Further, is there a particular reason to use `null` here, especially if the tests expects `undefined`?

mathieudutour: the reason to use `null` is that it gets translated to `nil` which is what obj-c expects. `undefined` is translated to `MOUndefined` because `undefined` just doesn't exist in Obj-C

PR #387, opened 2019-02-20
add `Unknown` to style type map

### LINKED ISSUE #385: A shared style from JSON returns styleType of 3
This test produces the problem: https://github.com/WCByrne/SketchAPI/blob/f701eb0b7bcd9de9d1cf0bbfa789329ee01b899e/Source/dom/__tests__/export.test.js#L74

mathieudutour:
fix #385

### DIFF
--- a/docs/api/SharedStyle.md
+++ b/docs/api/SharedStyle.md
@@ -109,9 +109,10 @@ You can unlink a Shared Style from the Library it comes from and make it a local
 SharedStyle.StyleType.Text
 ```
 
-Enumeration of the type of Shared Style.
+Enumeration of the type of Shared Style. Note that the `Unknown` shouldn't happen. It's a sign that the object is broken.
 
-| Value   |
-| ------- |
-| `Text`  |
-| `Layer` |
+| Value     |
+| --------- |
+| `Text`    |
+| `Layer`   |
+| `Unknown` |
--- a/docs/api/SharedStyle.md
+++ b/docs/api/SharedStyle.md
@@ -17,12 +17,12 @@ document.sharedTextStyles.push({
 
 A shared style (either a layer style or a text style).
 
-| Properties                                                                            |                                                   |
-| ------------------------------------------------------------------------------------- | ------------------------------------------------- |
-| id<span class="arg-type">string</span>                                                | The unique ID of the Shared Style.                |
-| styleType<span class="arg-type">[SharedStyle.StyleType](#sharedstylestyletype)</span> | The type of the Shared Style (`Layer` or `Text`). |
-| name<span class="arg-type">string</span>                                              | The name of the Shared Style.                     |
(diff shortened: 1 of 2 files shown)


bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/style/Style.js` changed, but not:
  - 📚 its docs
- 🧪 its tests

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

mathieudutour:
It happens in the case described in the issue: when exporting a shared style to JSON and re-importing it. The type of the shared style is encoded in its container, so it works fine when exporting the entire document. There might be some other cases but I don't know about them.

So i don't think it can be prevented, it's an edge case and shouldn't really happen in normal usage. That's why I didn't update the documentation, I don't really know how to convey that it shouldn't normally happen and that if it does happen, then I hope you know what you are doing 😄

christianklotz:
If it can't be prevented it should be documented. Two options spring to mind:

1. Make the `styleType` attribute optional, so it could result in an `undefined` and document the behaviour of the attribute.
2. Removing the mention of the `Image` and `Text` enum values from the `styleType` attribute docs and explicitly add `Unknown` to [`StyleType`](https://github.com/BohemianCoding/SketchAPI/blob/develop/docs/api/SharedStyle.md#sharedstylestyletype) with an explanation of when `Unknown` happens.

mathieudutour:
To be clear: I'm not the one that sets the `unknown` value, it's Sketch when it doesn't know how to deal with it. It shouldn't happen. It's a sign of a broken object and if the dev gets one of those, there is nothing we can do for them. I have no idea when it can happen, we just found one case but there could be others

christianklotz:
Aha, I see. Thank you for the clarification! In which case I suggest option 2, unless of course there's a possibility to fix this in Sketch to prevent `unknown` from ever happening?

docs/api/SharedStyle.md
```diff
@@ -109,9 +109,10 @@ You can unlink a Shared Style from the Library it comes from and make it a local
 SharedStyle.StyleType.Text
 ```
 
-Enumeration of the type of Shared Style.
+Enumeration of the type of Shared Style. Note that the `Unknown` shouldn't happen. It's a sign that the object is broken.
```
christianklotz: Let's remove _"(`Layer` or `Text`)"_ [above](https://github.com/BohemianCoding/SketchAPI/pull/387/files#diff-e332d3cb55f690441912233ba05205caR23) and leave the details to `SharedStyle.StyleType`.

> The type of the Shared Style (`Layer` or `Text`).

Also, "shouldn't happen" isn't the best choice of words, I suggest:

> Enumeration of the type of Shared Style. `Unknown` indicates the object is broken and Sketch can't determine the style type.

PR #388, opened 2019-02-20
do no use context as first document choice

### LINKED ISSUE #386: Document.getSelectedDocument does not point to current document since 53.1
Hi there,

I should clarify that this does not happen in all situations, rather in this specific one.

I have an action that fires on OpenDocument that bootstraps an instance of BrowserWindow (from sketch-module-webview). I then have a plugin command which accesses that instance by it's ID, and then shows it.

The simplified code looks like this:

```
const generateWindow = () => {
    const window = new BrowserWindow({
      identifier: 'my-identifier',
      show: false,
      ...
    });

    win.on('focus', () => {
      const document = Document.getSelectedDocument(); // always references the first opened document
    });
    return window;
}

export const onDocumentOpened = () => {
    generateWindow();
}

export default () => {
    const window = BrowserWindow.fromId('my-identifier');
    window.show();
}
```

The problem seems to be that any calls to Document.getSelectedDocument after the window is created, always point to the first document that was opened, so if any subsequent documents are opened then we are not able to select them.

Please let me know if you need any further clarification, but it seems to me that something related to the JS runtime inside Sketch has changed.

Thanks!
Ryen

mathieudutour:
fix #386

### DIFF
--- a/Source/dom/models/Document.js
+++ b/Source/dom/models/Document.js
@@ -25,6 +25,11 @@ export function getDocuments() {
 export function getSelectedDocument() {
   let nativeDocument
 
+  if (!nativeDocument) {
+    const app = NSDocumentController.sharedDocumentController()
+    nativeDocument = app.currentDocument()
+  }
+
   // skpm will define context as a global so let's use that if available
   if (typeof context !== 'undefined') {
     /* eslint-disable no-undef */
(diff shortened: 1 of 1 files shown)


bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  - 📚 its docs
- 🧪 its tests

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #404, opened 2019-03-07
Updating when overrides are populated

hellovai:
Overrides are null when an instance is created. Once an instance is pushed into the document, the overrides become available. This is ambiguous in the current documentation and might cost a lot of developer hours.


mathieudutour:
Thanks!

PR #409, opened 2019-03-08
Improve handling of export errors

WCByrne:
Wrap native export calls in try catch and convert native exceptions to more helpful error messages.

Converting the native errors seemed to be the easiest way to handle this.

See: https://sketch-friends.slack.com/archives/C5YKGT26B/p1552063120097100

### DIFF
--- a/Source/dom/export.js
+++ b/Source/dom/export.js
@@ -68,6 +68,24 @@ function exportToJSONFile(nativeObjects, options) {
   })
 }
 
+function convertNativeError(err) {
+  console.log(`err native ${err.nativeException}`)
+  if (err.nativeException) {
+    const nativeMessage = String(err.nativeException.description())
+    const missingParentFailures = [
+      'BCAssertion failed for expression: (request.rootLayer) != nil',
+      'An Obj-C exception occurred. Export Options must have an immutable doc specified',
+    ]
+    console.log(`native message ${nativeMessage}`)
+    if (missingParentFailures.indexOf(nativeMessage) >= 0) {
+      return new Error(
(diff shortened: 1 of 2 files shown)


bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/export.js` changed, but not:
  - 📚 its docs

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

Source/dom/export.js
```diff
@@ -84,12 +102,15 @@ function exportToImageFile(nativeObjects, options) {
     options
   )
 
-  // export the pages
-  pages.forEach(exporter.exportPage)
-
-  // export the layers
-  if (layers.length) {
-    exporter.exportLayers(layers)
+  try {
```
mathieudutour: I'm not too sure about wrapping everything in try/catch. That breaks any optimizations the compiler can make.

I'd rather explicitly check for it:
```js
if (!nativeObjects.parentPage()) {
  throw error
}
```

WCByrne: I started going that route but then realized there are some edge cases in there. For example, some cases don't need a parent. This seemed less likely to cause conflicts if the internals were to change (except the error messages of course but that is covered by the tests). 

Any idea what that difference in optimizations would be? Significant for this?

christianklotz:
Closing for now. May revisit at a later point.

PR #416, opened 2019-03-19
fix inserting shared styles

mathieudutour:
You can only insert local shared styles (eg. not linked to a Library). `document.sharedLayerStyles` returns the foreign shared styles (eg. linked to a Library) concatenated with the local shared styles.

So if you tried to insert a shared style (using `sharedStyles.push`), it would throw an index out of bounds error if there were some foreign shared styles.

Now it won't but if you try to insert a new Shared Style at the beginning (using `unshift` for example), it will end up at the beginning of the local Shared Styles but that might not be the beginning of all the shared styles if there are some foreign.

I don't think there is a perfect solution :(

For that reason, I decided, to de-deprecate `SharedStyle.fromStyle` which is just cleaner when you want to create a shared style and get a reference to it.

### DIFF
--- a/Source/dom/models/Document.js
+++ b/Source/dom/models/Document.js
@@ -622,162 +622,133 @@ Document.define('gradients', {
   },
 })
 
-Document.define('sharedLayerStyles', {
-  array: true,
-  get() {
-    if (!this._object) {
-      return []
-    }
-    const documentData = this._getMSDocumentData()
-    return toArray(documentData.allLayerStyles()).map(wrapObject)
-  },
-  set(sharedLayerStyles) {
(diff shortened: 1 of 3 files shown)


bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  - 📚 its docs
- 🧪 its tests

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

Issue #408, opened 2019-03-08
Add access to converting layers to outlines


WCByrne:
There is currently no access to the underlying `layersByConvertingToOutlines`

WCByrne:
Implementation: https://github.com/WCByrne/SketchAPI/tree/add-convert-to-outlines
Will open a PR if this is desired

PR #431, opened 2019-03-25
Add convert to outlines

WCByrne:
Addresses #408

### DIFF
--- a/Source/dom/layers/Layer.js
+++ b/Source/dom/layers/Layer.js
@@ -113,8 +113,9 @@ export class Layer extends WrappedObject {
   /**
    * Return a new layer
    */
-  layerByConvertToOutlines() {
-    return this._object.layersByConvertingToOutlines()
+  layersByConvertingToOutlines() {
+    const converted = this._object.layersByConvertingToOutlines()
+    return toArray(converted).map(l => wrapNativeObject(l))
   }
 
   // @deprecated
--- a/Source/dom/layers/Layer.js
+++ b/Source/dom/layers/Layer.js
@@ -111,7 +111,7 @@ export class Layer extends WrappedObject {
   }
 
   /**
-   * Return a new layer
+   * Return one or more layers the were converted to outlines.
    */
   layersByConvertingToOutlines() {
     const converted = this._object.layersByConvertingToOutlines()
--- a/Source/dom/layers/Layer.js
+++ b/Source/dom/layers/Layer.js
@@ -110,6 +110,13 @@ export class Layer extends WrappedObject {
     })
   }
 
+  /**
+   * Return a new layer
+   */
+  layerByConvertToOutlines() {
+    return this._object.layersByConvertingToOutlines()
+  }
+
   // @deprecated
   localRectToParentRect(rect) {
     console.warn(
--- a/docs/api/Layer.md
+++ b/docs/api/Layer.md
@@ -34,6 +34,18 @@ A new identical layer will be inserted into the parent of this layer.
 
 A new [Layer](#layer).
 
+## Convert the Layer to Outlines
+
+```js
+const layers = layer.layersByConvertingToOutlines()
+```
+
+Convert this layer to outlines.
+
+### Returns
+
+An array of new [Layer](#layer) objects representing the outlines of this layer.
+
 ## Remove the Layer
 
 ```javascript
--- a/CHANGELOG.json
+++ b/CHANGELOG.json
@@ -1,5 +1,8 @@
 {
-  "unreleased": ["[New] Add `isSelected` method on a CurvePoint"],
+  "unreleased": [
+    "[New] Add `isSelected` method on a CurvePoint",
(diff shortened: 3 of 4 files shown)


bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/layers/Layer.js` changed, but not:
  - 📚 its docs

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

docs/api/Layer.md
```diff
@@ -34,6 +34,18 @@ A new identical layer will be inserted into the parent of this layer.
...
+
+```js
+const layers = layer.layersByConvertingToOutlines()
+```
+
+Convert this layer to outlines.
+
+### Returns
+
+An array of new [Layer](#layer) objects representing the outlines of this layer.
```
WCByrne: Might be nice to clarify when this has more than one layer. I'm still not sure when that would be.

mathieudutour:
Ah sorry, I didn't comment on this. 

It feels a bit weird to me to have this method on a `Layer` instance. Ideally, I'd prefer to have something like `Layer.fromOutlines(layer)` but I'm not sure where it would go. 

But you are right that it can sometimes create multiple layers (when a shape has multiple borders IIRC). So I'm still not sure

WCByrne:
My opinion: The current name and location (`layer.layersByConvertingtoOutlines`) is more clear than having it on `Layer. `Layer.fromOutlines(layer)` actually feels a little more awkward to me. If anything, maybe something like `outlineLayersFromLayer(layer)` but that still reads weird.

mathieudutour:
yeah, I meant more about having it as a class method (eg. more like a constructor) rather than an instance method

WCByrne:
Any more thoughts on naming for this?

WCByrne:
@mathieudutour Let me know if and how you want me to update this. Happy to do whatever you think makes the most sense.

mathieudutour:
Yeah sorry, I need to find some time to play with a few different namings. Sorry for the delay :/

WCByrne:
No worries, just remembered this had been sitting here for a while. This is an easy one to drop in to the internals for now.

iliakan:
Is there a way to convert to outlines for JS Sketch plugin right now?

Thank you.

WCByrne:
This PR needs to address this issue that was [brought up in Slack](https://sketch-friends.slack.com/archives/CCXDEQG9H/p1570219535048800). @christianklotz any insight on how this can be achieved in code? I'd be happy to implement and get this PR moving again.

> It looks like converting text to outlines is filling in letters in unexpected ways. I believe this started with 58 according my user’s reports.
> 
> image
> 
> **Response:** altering the fill type from Non-Zero to Even-Odd will likely achieve the intended result there (via the cog button in the Fills section within the Inspector)

christianklotz:
This is going to be fixed in Sketch 60. Once out it'd be a good time to pick up this PR again.

Source/dom/layers/Layer.js
```diff
@@ -110,6 +110,14 @@ export class Layer extends WrappedObject {
     })
   }
 
+  /**
+   * Return one or more layers the were converted to outlines.
+   */
+  layersByConvertingToOutlines() {
```
christianklotz, line 116:
```suggestion
  toOutlines() {
```

Maybe `toOutlines`, similar to `String.toUppercase` would suffice?

christianklotz (commented):
The issue here is that this would actually fail on any layer that doesn't support conversion to outlines, e.g. symbol instances, groups, etc. There's a method `canConvertToOutlines` which returns true for anything that can be converted, posing the question of how this would be translated here. Does it only return itself wrapped in an array or `null`/`undefined`?

Source/dom/layers/Layer.js
WCByrne: Makes sense to me.

WCByrne:
As for failing when `canConvertToOutlines` is false, do the internals return nil or itself? Should that check function be adde here as well?

christianklotz:
@WCByrne it gets never called because it's only implemented for layer subclasses that support it.

WCByrne:
got it. I feel like returning nil then, or maybe even throw an error to make it clear why it didn't work. If you are trying to convert something to outlines, getting that layer back in an array seems like it could be confusing

christianklotz:
Returning nil works for me. Throwing an error is a bit harsh, especially if the method is on `Layer`. An alternative could be to put it on `Text` and `ShapePath` only?

WCByrne:
If those are the only 2 that support it, that works for me. I was just kinda going with what the internals seemed to do. FWIW, I only use it for text layers so this is a bit of a none issue for my needs.

christianklotz:
👍 … feel free to make the change so only the two classes have the method and I'll review.

PR #451, opened 2019-04-04
fix printing the prototype of a wrapped object

### LINKED ISSUE #449: Why is there no isSymbolsPage property on page?
Sketch 54 adds a method on `Page` called `isSymbolsPage()`. I'm curious why this is a method rather than a property. This leads me another question. How can I get all the methods for an object? In javascript land I would do `object.prototype`

mathieudutour:
Fix #449 

To check if an object is a wrapped object, we check for `object. _isWrappedObject` so we need to define that property only on an instance, and not on the prototype (because the prototype isn't a wrapped object)


bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/dom/WrappedObject.js` changed, but not:
  - 🧪 its tests

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #460, opened 2019-04-10
Code Review: Restrict url scheme to launch a plugin to a specific command (#24181)

mathieudutour:
Resolve BohemianCoding/Sketch#24181


WCByrne:
This is great! Just for clarification: 
1. Does this mean that the url _must_ include a command now? 
2. Will the HandleURL actionable still be called?

mathieudutour:
1. No but if it doesn't, your command won't be called (so kind of yes)
2. When the URL is like `sketch://plugin/my.plugin.identifier/my.command.identifier`, yes

wuyingfeng1992:
The url can in terminal execute and skip the open sketch notify？

PR #467, opened 2019-04-13
Kevin/fix custom shape type documentation

KevinGutowski:
I fixed the documentation and attempted to fix the test for `ShapePath` but I’m a little confused with the eslint constraint with enforcing property shorthands. When I copy the code into the Script Panel the shapePath object isn’t created. Perhaps the tests get compiled into something that Sketch can properly execute.

I am also a little confused about this comment under `ShapePath.js`

```
const ShapeTypeMapReverse = {
  Rectangle: MSRectangleShape,
  Oval: MSOvalShape,
  Polygon: MSPolygonShape,
  Star: MSStarShape,
  Triangle: MSTriangleShape,
  Custom: MSRectangleShape, // we are just going to default to Rectangle here
}
```
This seems to be the root of the problem. Why isn’t `Custom` mapped to `MSShapePathLayer`? This means you can’t create lines programmatically in the same way that you can with the line tool.

This PR needs more work but hopefully its helpful in pointing out the areas of issues.

### DIFF
--- a/docs/api/ShapePath.md
+++ b/docs/api/ShapePath.md
@@ -32,11 +32,11 @@ A shape path layer. It is an instance of [Layer](#layer) so all the methods defi
 ```javascript
 const shapePath = new ShapePath({
   name: 'my shape path',
-  shapePath: ShapePath.ShapeType.Oval,
+  shapeType: ShapePath.ShapeType.Oval,
 })
 ```
 
-You can only set the `shapePath` when creating a new one. Once it is created, the `shapePath` is read-only. If it is not specified, it will default to ShapePath.ShapeType.Rectangle
+You can only set the `shapeType` when creating a new one. Once it is created, the `shapeType` is read-only. If it is not specified, it will default to ShapePath.ShapeType.Rectangle
 
 ```javascript
 const shapePath = ShapePath.fromSVGPath('M10 10 H 90 V 90 H 10 L 10 10')


KevinGutowski:
I see two ways of going about fixing this. Adding a new `ShapeType` called “Line” or modifying `ShapeTypeMapReverse`

mathieudutour:
Hum. I can see the issue. It's default to `Rectangle` for backward compatibility. But maybe when points are specified when creating a Shape, we shouldn't default to Rectangle.

Adding `ShapeType.Line` is tricky because Sketch will always say that it's Custom. So we would need to construct the type by ourselves in the API (we could count the points).

KevinGutowski:
Gotcha, yeah it seems like I’m missing something for how to create a line. I even tried something like this and it doesn’t behave like the line tool:

```
let shapePathLayer = MSShapePathLayer.alloc().init()
let myLine = sketch.fromNative(shapePathLayer)
console.log(myLine.shapeType)
// 'Custom'
myLine.name = 'myLine'
myLine.frame = new Rectangle(0,0,40,100)
myLine.style = { borders: ['#FF0000'] }
myLine.points = [{ point: { x: 0, y: 0 }},{ point: { x: 1, y: 1 }}]
myLine.parent = myArtboard
```

I’m not sure what type of object the line tool is making exactly 🤔 . How about for now, I'll rename this PR and just fix the documentation. I’ll also add some more details to #467 about what is going on.

mathieudutour:
Thanks for the documentation change. Could you run `git reset --hard HEAD~2` in your branch to remove the last 2 commits and then run `git push --force`?

KevinGutowski:
Done!

PR #475, opened 2019-04-24
Add Textarea option to UI.getInputFromUser

### LINKED ISSUE #450: Support multiline input in UI.getInputFromUser
It would be great for this plugin I wrote, where a single line input is clearly not cutting it : )

Screen Shot 2019-04-04 at 10 34 44

KevinGutowski:
fixes #450

Tried following along from
- https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/TextUILayer/Tasks/TextInScrollView.html
- https://github.com/AnimaApp/CocoaPowder/blob/788cbfad1a2abb93fd0a3c5bbeb77b8ebd8d3d6f/CocoaPowder%20Sample.sketchplugin/Contents/Sketch/CocoaPowder.js#L123

Todos:
- [x] Figure out how to appease linter?
- [x] Figure out what the frame height for the accessory view should be so that the multiplier works properly
- [x] Update Documentation

### Example usage
```
UI.getInputFromUser("What's your favorite design tool?", {
  type: UI.INPUT_TYPE.textarea,
  numberOfLines: 10, // defaults to 3,
  initialValue: 'hi',
}, (err, value) => {
  if (err) {
    // most likely the user canceled the input
    return
  }
  console.log(value)
})
```

### Screenshot
Screen Shot 2019-04-24 at 1 07 14 AM


Source/ui/UI.js
```diff
@@ -92,6 +93,48 @@ export function getInputFromUser(messageText, options, callback) {
       )
       dialog.window().setInitialFirstResponder(accessory)
       break
+    case INPUT_TYPE.textarea:
```
mathieudutour: Do you think it's necessary to add another input type? I was thinking of just having it as having the `numberOfLines` option default to 1  and if it's different than 1, then change the text input to be a textarea

KevinGutowski: Ah fair point. That is probably easiest since the other input type is `string` and not `textfield`. Though both on the web and on macOS those are pretty explicitly two different objects.

KevinGutowski: Moved it over into the `string` type. Seems like the switch statement is getting pretty big now!

mathieudutour: I think that's fine, we can refactor into smaller functions when we need it

mathieudutour:
To "appease" the linter, you will need to add a few globals to the `.eslintrc.yml`: `NSScrollView`, `NSViewWidthSizable`, `NSTextView`, `NSMakeSize` and `NSViewWidthSizable `.  You will also need to add `// eslint-disable-next-line no-bitwise` in line 95

bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  `Source/ui/UI.js` changed, but not:
  - 🧪 its tests

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

KevinGutowski:
@mathieudutour When you get a chance, let me know if there is anything else I can do to help this PR along!

Issue #421, opened 2019-03-21
Improve documentation


christianklotz:
The documentation has grown over time and while very detailed in many places it is missing content in others and makes it difficult to find certain information. Another aspect that's been brought up multiple times is a better documentation of changes, see #140, #420 and #411.

This is an umbrella issue for everything related to restructuring and updating the documentation content. Below is a work in progress structure which is split into four sections; the Sketch app, the command line interface, plugin development and the file format.

Each section follows the same structure, introducing wider _Concepts_ and optionally providing _Guides_ for real world use cases which allows for a range of entry level to more advanced content.

### Sketch Developer
- Sketch app
  - Using the sketch:// URL scheme
- `sketchtool` CLI
  - Concepts
    - Versioning
    - Usage
  - Export
    - Pages
    - Artboards
    - Layers
    - Slices
    - Preview
  - Run plugins
  - List document structure
  - Show locations
  - Extract metadata
  - Dump file into JSON
- Plugin
  - Guides
    - Run Script within Sketch
    - Install a plugin
    - Remove a plugin
    - Create a Plugin
    - Debug a plugin
    - Publish a Plugin using skpm
    - Publish a Plugin manually
    - Update a Plugin
    - Handle an Action
    - Use an external text editor
    - Add a native user interface
    - Add a user interface using a webview
  - Concepts
    - Plugin Bundle
    - Sketch JavaScript Environment
    - Manifest
    - Sketch menu
    - Commands
    - Actions
    - API Versioning
  - JavaScript API
    - DOM
    - UI
    - Data
    - Settings and State Preservation
    - Actions
    - Changelog
  - Internal API
    - Disclaimer
    - CocoaScript
    - Sketch Headers
- File format
  - Concepts
    - Versioning
    - File Bundle
  - Sample file
- API Reference

christianklotz:
@KevinGutowski I noticed your questions on the [forum](https://sketchplugins.com), your efforts to create some sample code and it sounded like you were planning to put things together in a PR. If this is the case it may be good to coordinate what I'm doing with the stuff you're working on.

As you can see in ☝️ content outline I'm primarily doing two things:

- rearranging content
- revisiting and producing new concept and guide pages for the different sections (plugins, CLI, file format)

What I'm not doing at this stage is looking at the content of the actual API reference. It feels like the work you've been doing is primarily related to the reference. Is this correct? If true we'd have updates in both areas which would be fantastic 🤗

robintindale:
[...]

PR #476, opened 2019-04-24
add docs for handlers and manifest schema

mathieudutour:
I realised that the `handlers` field isn't documented and I always forget the manifest syntax anyway so I created a json schema.

The schema is automatically picked up by vscode when referencing it with a `$schema` property:
image

There is the corresponding PR on skpm to add the schema reference: https://github.com/skpm/skpm/pull/223

_I'm guessing the doc could be generated from the schema but that can wait_


christianklotz:
Holding this one until #421 is completed

Issue #507, opened 2019-05-21
Add What's new in Sketch 55 based on Kevin's post on Sketch Plugins


christianklotz:
https://sketchplugins.com/d/1394-what-s-new-sketch-beta-55

PR #508, opened 2019-05-21
Add 55 changes including separate section in URL scheme

### LINKED ISSUE #506: Add new URL scheme for running a plugin command to documentation
```
sketch://plugin/my.plugin.identifier/my.command.identifier
```

christianklotz:
Closes #506 and #507

### DIFF
--- a/docs/pages/app.md
+++ b/docs/pages/app.md
@@ -55,6 +55,7 @@ The command must implement the `HandleURL` action and have it defined in the plu
 ```js
 const sketch = require('sketch')
 
+// If you're using skpm: export function handleURL(context)
 function handleURL(context) {
   let query = context.actionContext.query
   sketch.UI.message(query.msg || '👋')
--- a/docs/pages/app.md
+++ b/docs/pages/app.md
@@ -14,7 +14,9 @@ You can integrate Sketch into your app or worflow using the `sketch://` URL sche
 
 The URL scheme accepts two parameters. If the path to the file is provided without any parameters the file will be opened in the same way it was last saved.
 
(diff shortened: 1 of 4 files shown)


docs/pages/app.md
```diff
@@ -33,3 +35,54 @@ document.selectedLayers.forEach(layer => console.log(layer.id))
...
+#### Example
+
+The command must implement the `HandleURL` action and have it defined in the plugin manifest in order to be run.
+
+##### `command.js`
+
+```js
+const sketch = require('sketch')
+
+function handleURL(context) {
```
mathieudutour: a bit tricky here: if using skpm, this will not work, it needs to be `export function handleURL...`

christianklotz: Hmm, yeah … I stumbled into this, too. Perhaps I just add a comment above?

```js
// export function handleURL(context) if you're using skpm
```

mathieudutour: sounds good

docs/pages/plugins-releases-55.md
```diff
@@ -0,0 +1,199 @@
...
+})
+```
+
+### Some better logging of the prototype of wrapped objects
+
+[_GitHub pull request_](https://github.com/BohemianCoding/SketchAPI/pull/451)
+
+There was a bug in the `util.inspect` algorithm (which console uses) that makes it think the prototype of a wrapped object is a wrapped object and uses the code path for wrapped object. This has been fixed for better logging output.
+
+### Upcoming Additions
```
mathieudutour: not sure we should put this here, it will be for 56 ;)

christianklotz: Yeah, ok. Let's take it out

Issue #514, opened 2019-05-23
Add coordinates concept to file format page


nthState:
Explain the coordinate system used by the SketchAPI/Open file format.

Add a concept section to https://developer.sketch.com/file-format/ page on the developer docs.

Examples to show:
- Coordinate usage in the API vs Open File Format
- Results when nested
- Results when rotated

PR #517, opened 2019-05-24
Added page on normalized coordinates

nthState:
Added a page describing the normalized values saved to a sketch document, linked to issue https://github.com/BohemianCoding/SketchAPI/issues/514


docs/pages/file-format-coordinates.md
```diff
@@ -0,0 +1,105 @@
...
+title: Coordinate Systems
+section: file-format
+chapter: Concepts
+permalink: /file-format/coordinate-systems
+
+order: 102
+excerpt: How Sketch stores coordinate information in page.json
+---
+
+When creating shapes, whether it's in Sketch, or in the API, you use define sizes naturally.
```
mathieudutour: The entire page is only about the points of a path of a shape but it's only mentioned near the bottom of the page. 

Also when you defines the points of a ShapePath in the API, you need to use the normalised coordinates, not what you call the natural sizes.

jedrichards: Typo here?

> you use define sizes naturally

docs/pages/file-format-coordinates.md
```diff
@@ -0,0 +1,105 @@
...
+| x: 40, y: 0  | 40 / frame.width, 0 / frame.height | 1, 0 |
+|--------------|---------------------------------|-----|
+| x: 40, y: 40 | 40 / frame.width, 40 / frame.height | 1, 1 |
+|--------------|---------------------------------|-----|
+| x: 0, y: 40  | 0 / frame.width, 40 / frame.height | 0, 1 |
+|--------------|---------------------------------|-----|
+
+Note: JSON has been snipped for brevity.
+
+```json
```
jedrichards: Can we format this JSON as if pretty printed? The whitespace looks a bit wonky atm.

mathieudutour: @nthState if you run `npm install`, it should install a pre-commit hook that format the markdown for you. Depending on which editor you use, you can also use a Prettier extension to format on save. (and if you use vs code, there should be extensions suggestions)

docs/pages/file-format-coordinates.md
nthState: @mathieudutour - I've updated the page name/contents to reflect this

mathieudutour: err, I think you removed the page completely haha

nthState: :-) It's back now

PR #526, opened 2019-05-29
Improve manifest.json docs

jedrichards:
I tripped up frequently trying to create a vanilla plugin using the existing manifest.json documentation. Specifically how the manifest interacts with how functions should be declared in a vanilla plugin vs skmp, and the differences between `hanlder` and `handlers`. This is my attempt to improve it for new users.

### DIFF
--- a/docs/pages/plugins-create-plugin.md
+++ b/docs/pages/plugins-create-plugin.md
@@ -116,7 +116,7 @@ To verify the plugin has been updated, create or open a Sketch document and sele
 
 ## 2. Manual
 
-If you prefer using your own method to bundle the plugin, you must make sure to define the manifest correctly and include all scripts the required for the plugin to run.
+If you prefer using your own method to bundle the plugin, you must make sure to define the manifest correctly and include all scripts required for the plugin to run.
 
 ### Example
 
@@ -184,6 +184,14 @@ const randomEmoji = () => {
 module.exports = {
   randomEmoji,
 }
+```
+
+#### Running the plugin
+
+You can use [`sketchtool`](/cli) to run the plugin within Sketch from the command line.
+
+```
+sketchtool run ./plugin.sketchbundle messages.greeting
 ```
 
 ## Next steps
--- a/docs/pages/plugins-create-plugin.md
+++ b/docs/pages/plugins-create-plugin.md
@@ -188,11 +188,9 @@ module.exports = {
 
 #### Running the plugin
 
-You can use [`sketchtool`](/cli) to run the plugin within Sketch from the command line.
+Place your plugin in the [plugins location](/plugins), and once located there you can run the plugin manually from the plugins menu within Sketch.
 
-```
-sketchtool run ./plugin.sketchbundle messages.greeting
-```
+> Alternatively, if you're comfortable on the command line you can use [`sketchtool`](/cli) to run the plugin instead.
 
 ## Next steps
 
--- a/docs/pages/plugins-manifest.md
+++ b/docs/pages/plugins-manifest.md
@@ -10,7 +10,7 @@ excerpt: The manifest is a JSON file that provides information about a plugin su
 
 The manifest is a JSON file that provides information about a plugin such as author, description, icon and where to get the latest update from. It also tells Sketch how to run the plugin and how it integrates into the _Plugins_ menu.
 
-> If you're interested in validating a `manifest.json` file, or understanding the file structure in detail, we maintain a JSON Schema [here](https://github.com/BohemianCoding/SketchAPI/blob/615874d7a3d12a00b56fcbabdf1a568ef7bb0ad0/docs/sketch-plugin-manifest-schema.json)
+> **Tip:** Validate your `manifest.json` using the [JSON Schema](https://github.com/BohemianCoding/SketchAPI/blob/develop/docs/sketch-plugin-manifest-schema.json).
 
 ## Example manifest
 
--- a/docs/pages/plugins-manifest.md
+++ b/docs/pages/plugins-manifest.md
@@ -10,11 +10,11 @@ excerpt: The manifest is a JSON file that provides information about a plugin su
 
 The manifest is a JSON file that provides information about a plugin such as author, description, icon and where to get the latest update from. It also tells Sketch how to run the plugin and how it integrates into the _Plugins_ menu.
 
+> If you're interested in validating a `manifest.json` file, or understanding the file structure in detail, we maintain a JSON Schema [here](https://github.com/BohemianCoding/SketchAPI/blob/615874d7a3d12a00b56fcbabdf1a568ef7bb0ad0/docs/sketch-plugin-manifest-schema.json)
+
 ## Example manifest
 
-The example plugin below called _Select Shapes_ defines three commands _All_, _Circles_ and _Rectangles_ which can be accessed from _Plugins_ › _Select Shape_.
-
-It also includes an `appcast` URL for Sketch to automatically check if a new version of the plugin is available.
+The example `manifest.json` file below is for a plugin called _Select Shapes_, and defines three commands - _All_, _Circles_ and _Rectangles_. These commands would be accessible from the _Plugins_ › _Select Shape_ menu in Sketch.
 
 ```json
 {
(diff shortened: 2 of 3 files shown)


docs/pages/plugins-manifest.md
```diff
@@ -81,55 +81,78 @@ Defines the minimum version of Sketch required to run the plugin. This string mu
...
 
+The function in `selection-changed.js` would be written as,
+
 ```js
 // onSelection handler
-export function onSelection(context) {
-  var doc = context.document;
-  var selection = context.selection;
-  …
+function onSelection(context) {
```
mathieudutour: so I think we need to be consistent in the docs. All the introduction is based on skpm because it's a lot easier to setup. So either we consider skpm the default, and the examples should use `export function`, or we need to rewrite the introduction

christianklotz: This is not quite true. When I rewrote the documentation I assumed vanilla implementation to be default and when `skpm` is mentioned, it's specifically a provided as a convenience.

mathieudutour: True, but in the guides, skpm is always the first option (for "Create a plugin", "Publish a plugin", "update a plugin", etc.)

christianklotz: That's right, but `skpm` is specifically mentioned in these guides. It might be a good reason to add a litte note in the _Create a plugin_ guide making it obvious because that's the only place where a handler is mentioned.

docs/pages/plugins-manifest.md
```diff
@@ -10,11 +10,11 @@ excerpt: The manifest is a JSON file that provides information about a plugin su
...
 The manifest is a JSON file that provides information about a plugin such as author, description, icon and where to get the latest update from. It also tells Sketch how to run the plugin and how it integrates into the _Plugins_ menu.
 
-## Example manifest
+> If you're interested in validating a `manifest.json` file, or understanding the file structure in detail, we maintain a JSON Schema [here](https://github.com/BohemianCoding/SketchAPI/blob/615874d7a3d12a00b56fcbabdf1a568ef7bb0ad0/docs/sketch-plugin-manifest-schema.json)
 
-The example plugin below called _Select Shapes_ defines three commands _All_, _Circles_ and _Rectangles_ which can be accessed from _Plugins_ › _Select Shape_.
+## Example manifest
 
-It also includes an `appcast` URL for Sketch to automatically check if a new version of the plugin is available.
+The example `manifest.json` file below is for a plugin called _Select Shapes_, and defines three commands - _All_, _Circles_ and _Rectangles_. These commands would be accessible from the _Plugins_ › _Select Shape_ menu in Sketch.
```
christianklotz: Two minor formatting issues:

- Instead of "…three commands - _All_…": "…three commands – _All_…" (replace hyphen) 
- ~_Select Shape_~ _Select Shapes_ menu in Sketch

docs/pages/plugins-manifest.md
```diff
@@ -81,55 +81,78 @@ Defines the minimum version of Sketch required to run the plugin. This string mu
...
+| `handler`     | Specifies the name of function to be called with the command. The function should be declared at the top level of the script, and accepts a single `context` parameter which contains information such as the current document and selection.<br/><br/>If using the `skpm` tool you may `export` the function instead of declaring it, or omit this field entirely and declare the handler with `export default` |
+| `handlers`    | If more fine-grain control is required use the `handlers` field instead of `handler`, see below for in-depth documentation                                                                                                                                                                                                                                                                                       |
 
 ##### `handlers`
 
-An object used to specify the mapping between events (like Actions, calling the command, etc.) and function names.
+An object used to fully specify a command's _setup > run > tearDown_ lifecycle, as well as work with [Actions](/guides/action-api/). When specified the `handler` field is ignored.
```
christianklotz: Suggestion for improved formatting:

_setup_ › _run_ › _tearDown_

or 

`setup › run › tearDown`

docs/pages/plugins-manifest.md
jedrichards: Agreed consistency is key. As an end-user to the docs it wasn't clear to me which one was the default usage pattern. If I just arrive on the manifest.json page for example, I have no idea whether skpm is the first option on other pages or not, so that's not helpful. It's awkward that there are two competing approaches to creating a plugin, and while that's the case I think the documentation policy _has_ to always make it abundantly explicit which one is being discussed at every point.

docs/pages/plugins-manifest.md
```diff
@@ -81,55 +81,78 @@ Defines the minimum version of Sketch required to run the plugin. This string mu
 
 An array of objects defining all commands provided by the plugin.
 
+For example, the example command definition below says that the `selectAll` JS function in `script.js` should be called when the `All` command is invoked from Sketch.
```
mathieudutour:
```suggestion
For example, the example command definition below says that the `selectAll` JS function in `script.js` should be called when the `Select All` command is invoked from Sketch.
```

docs/pages/plugins-manifest.md
```diff
@@ -81,55 +81,78 @@ Defines the minimum version of Sketch required to run the plugin. This string mu
...
+| `script`      | Specifies the relative path within the plugin bundle's `Sketch` folder to the script implementing the command.                                                                                                                                                                                                                                                                                                   |
+| `handler`     | Specifies the name of function to be called with the command. The function should be declared at the top level of the script, and accepts a single `context` parameter which contains information such as the current document and selection.<br/><br/>If using the `skpm` tool you may `export` the function instead of declaring it, or omit this field entirely and declare the handler with `export default` |
```
mathieudutour: if you are not using skpm and omit the field, you need to declare an `onRun` function

docs/pages/plugins-manifest.md
```diff
@@ -81,55 +81,78 @@ Defines the minimum version of Sketch required to run the plugin. This string mu
...
-  var doc = context.document;
-  var selection = context.selection;
-  …
+function onSelection(context) {
+  var doc = context.document
+  var selection = context.selection
 }
 ```
 
+> Functions referenced in the `handlers` object are defined in the same was as the `handler` function, i.e. declared at the top level of the script in a vanilla plugin, or using ES6 module `export` syntax with `skpm`.
```
mathieudutour:
```suggestion
> Functions referenced in the `handlers` object are defined in the same way as the `handler` function, i.e. declared at the top level of the script in a vanilla plugin, or using ES6 module `export` syntax with `skpm`.
```

docs/pages/plugins-manifest.md
christianklotz: If vanilla is default it would need to be worded the other way round:

If you're using `skpm` you don't need to declare an `onRun` function.

docs/pages/plugins-manifest.md
jedrichards: Oh man, I almost used an emdash too and thought surely Chrisitian won't notice that 😬

docs/pages/plugins-create-plugin.md
```diff
@@ -186,6 +186,14 @@ module.exports = {
 }
 ```
 
+#### Running the plugin
```
mathieudutour: I'm not sure we want to introduce sketchtool when a user just create a plugin. 99% of the users will just run the plugin via the Sketch Plugin menu

christianklotz: Agree with @mathieudutour here, the majority will run the menu through the menu. However there could be a note box:

> *Note:* To run a plugin from the command line you can use [`sketchtool`](/cli)

… or something along the lines

docs/pages/plugins-manifest.md
```diff
@@ -10,11 +10,11 @@ excerpt: The manifest is a JSON file that provides information about a plugin su
 
 The manifest is a JSON file that provides information about a plugin such as author, description, icon and where to get the latest update from. It also tells Sketch how to run the plugin and how it integrates into the _Plugins_ menu.
 
-## Example manifest
+> If you're interested in validating a `manifest.json` file, or understanding the file structure in detail, we maintain a JSON Schema [here](https://github.com/BohemianCoding/SketchAPI/blob/615874d7a3d12a00b56fcbabdf1a568ef7bb0ad0/docs/sketch-plugin-manifest-schema.json)
```
christianklotz: Perhaps more succinct and also linking to the main branch?

> **Tip:** Validate your `manifest.json` using the [JSON Schema](https://github.com/BohemianCoding/SketchAPI/blob/develop/docs/sketch-plugin-manifest-schema.json).

jedrichards: Ah, OK.

Issue #525, opened 2019-05-29
Add additional reference files


jedrichards:
As part of #515, flesh out the reference file section with addition types. Specifically:

- [x] Empty (Done in #516)
- [x] Shapes (#528)
- [x] Symbols (#528)
- [x] Foreign symbols and styles, e.g. Libraries (#532)
- [x] Symbol overrides (#536)
- [x] Fonts, text and typography (#536)
- [x] Groups (#537)
- [x] Images bitmap and vectors (#539)
- [x] Prototyping (#545 )

PR #528, opened 2019-05-31
 Add reference file automation scripts, and shape and symbol reference plugins

jedrichards:
Refs #525


mathieudutour:
Instead of creating 1 plugin for each file, perhaps it'd be easier to maintain 1 plugin with 1 command for each file? Also it'd be interesting to pretty print the JSON files

jedrichards:
@mathieudutour Agreed on both counts. Tasks for the next PR maybe?

jedrichards:
@mathieudutour Added some scripts to automate generating the files, and also added pretty printing the JSON and moved to a single plugin with multiple commands, as suggested. See what you think ...

scripts/generate-reference-files-md.js
```diff
@@ -0,0 +1,58 @@
...
+const { readdirSync } = require('fs')
+const table = require('markdown-table')
+const { writeFileSync } = require('fs')
+const { execSync } = require('child_process')
+
+const dir = resolve(__dirname, '../reference-files')
+const pages = resolve(__dirname, '../docs/pages')
+const pluginPath = resolve(dir, './plugin.sketchplugin')
+const manifest = require(resolve(pluginPath, 'Contents/Sketch/manifest.json'))
+const repoPath =
```
mathieudutour: I think you can take `https://github.com/BohemianCoding/SketchAPI` from the package.json

reference-files/plugin.sketchplugin/Contents/Sketch/shapes.js
```diff
@@ -26,7 +26,7 @@ function main(ctx) {
   })
 
   // Save and close
-  doc.save(ctx.pwd + '/output.sketch', {
+  doc.save(String(ctx.savePath), {
     saveMode: Document.SaveMode.SaveAs,
   })
   doc.close()
```
mathieudutour: I missed that earlier but the `.close()` should be in the callback of the `.save` method

scripts/generate-reference-files-md.js
```diff
@@ -0,0 +1,58 @@
...
+        cmd.identifier
+      }.js)`,
+    ]
+  }),
+])}`
+
+  const mdPath = resolve(pages, `file-format-reference-${cmd.identifier}.md`)
+
+  writeFileSync(mdPath, md, 'utf8')
+  execSync(`npm run prettier -- --write "${mdPath}"`, { encoding: 'utf8' })
```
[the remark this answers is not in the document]
jedrichards: Ah of course, `npm run` will pick up prettify automatically, good catch.

scripts/generate-reference-files.js
```diff
@@ -0,0 +1,63 @@
...
+ *   This script expects a working version of sketchtool to be in the path
+ *   and pointing at the right Sketch version. No Sketch version checking is
+ *   performed.
+ */
+
+const { execSync } = require('child_process')
+const del = require('del')
+const { resolve } = require('path')
+
+const sketchVersion = process.argv[2] || '55'
```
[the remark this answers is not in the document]
jedrichards: Ok roger.

scripts/generate-reference-files-md.js
jedrichards: Yep.

reference-files/plugin.sketchplugin/Contents/Sketch/shapes.js
jedrichards: Ah ok, yes makes sense.

scripts/generate-reference-files.js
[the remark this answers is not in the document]
jedrichards: The `sketchtool` output isn't exactly nicely machine readable:

```
sketchtool Version 55.1 (78136)
```

Easy enough to parse out though, unless there's another way of getting the Sketch version in a script?

scripts/generate-reference-files-md.js
jedrichards: Hmm, the `respository.url` field in package.json is for urls to the actual `.git` repo, not the http location of the GitHub web ui. Going to add the `homepage` field to package.json, and grab from there ...

jedrichards:
> One thing tho is that you are using sketchtool directly

Not too concerned about this at the moment, since I mention it in the script comments, and I'm guessing it'll mainly be me running the script for the time being! Your approach of passing in the path sounds good as a future improvement though

> Another thing: it'd be nice to change the UUID of the generated files to a constant string so that diffs are easier to read.

What would be the approach here? Manipulate the generated JSON itself, or is there a way to use specific/hard coded uuids through the API? I'm not sure I'd like to add too much code to the script that was dependant on particular a data structures/field name in the JSON if I can get away with it ...

mathieudutour:
> What would be the approach here?

I was thinking about a simple regex to run on the JSON file: `/"do_objectID": "([A-D0-9\-]+)"/`. There is no support for a hardcoded uuid in the API because it's usually a bad idea to change the id of sketch objects (Sketch is going to be really confused if you change certain IDs)

PR #532, opened 2019-06-05
Library related ref files

jedrichards:
Refs #525


jedrichards:
@mathieudutour Okay, ready for a review now 👀

jedrichards:
Hmm. Not sure ... we provide a link to them on the each doc page, as in https://develop--developer-sketch.netlify.com/file-format/reference/empty

The thinking being that it could be useful for somebody to open them up and have a poke around in the Sketch app to understand what's going on better?

mathieudutour:
Ah ok. I'm still not really comfortable putting binary files in an active code repo. Maybe all this generated files should go in a different repo. I'm still really linking the idea of having the files on branches instead of nested folders. Having nested folders makes it impossible to see diffs on github

jedrichards:
Yep agreed. My feeling is that the developer docs website probably belongs in its own repo entirely anyway. After all, the developer website relates to the whole developer platform, not just the JS API. I think generating the static site from this repo over complicates things.

There are much nicer static site generators available now too (e.g. Gatsby etc) that would the developer site more approachable for updates/fixes/contributions from the community. Also if we generated the developer site with a framework like Gatsby it'd make it easier to add rich features ... like rendering diffs between files etc.

mathieudutour:
> My feeling is that the developer docs website probably belongs in its own repo entirely anyway. After all, the developer website relates to the whole developer platform, not just the JS API. I think generating the static site from this repo over complicates things.

That's debatable and completely unrelated 😄. Debatable because you really want to keep the documentation next to the code. And using Jekyll or Gatsby doesn't really change anything. We'd never be able to build a nicer UI that what GitHub is doing for showing diffs, that's the whole point of GitHub. So I'd say we should leverage it instead of investing a lot into a sub-par UI.

jedrichards:
Well, let's stick to the current approach for now until the next planning session and we can discuss how/if to iterate the approach. I think it's good enough for now as the MVP of this feature, and easy to change since its mostly automated with the script.

PR #533, opened 2019-06-06
Add api for exposing and changing a document's color space

### LINKED ISSUE #531: Add support for Document's color profile
A Document can have a different color profiles (sRGB, P3, auto). We need to expose it in the API.

jedrichards:
Closes #531

### DIFF
--- a/Source/dom/models/Document.js
+++ b/Source/dom/models/Document.js
@@ -16,10 +16,16 @@ export const SaveModeType = {
   SaveAs: NSSaveAsOperation,
 }
 
-export const ColorSpaceType = {
+export const ColorSpaceMap = {
   Unmanaged: 0,
   sRGB: 1,
   P3: 2,
+}
+
+export const ColorSpace = {
+  Unmanaged: 'Unmanaged',
+  sRGB: 'sRGB',
+  P3: 'P3',
 }
 
 /* eslint-disable no-use-before-define, typescript/no-use-before-define */
@@ -380,6 +386,26 @@ export class Document extends WrappedObject {
 
     msdocument.close()
   }
+
+  changeColorSpace(colorSpace, convert = false) {
+    if (this.isImmutable()) {
+      return
+    }
+
+    const targetColorSpace = ColorSpaceMap[colorSpace]
+    const currColorSpace = this._getMSDocumentData().colorSpace()
+
+    if (typeof targetColorSpace === 'undefined') {
+      throw new Error(`Invalid colorSpace ${colorSpace}`)
+    }
+
+    // If the current colorSpace is unmanaged then we always assign
+    if (!convert || currColorSpace === ColorSpaceMap.Unmanaged) {
+      this._getMSDocumentData().assignColorSpace(targetColorSpace)
+    } else {
+      this._getMSDocumentData().convertToColorSpace(targetColorSpace)
+    }
+  }
 }
 
 Document.type = Types.Document
@@ -395,7 +421,8 @@ if (typeof MSDocument !== 'undefined') {
 }
 
 Document.SaveMode = SaveModeType
-Document.ColorSpace = ColorSpaceType
+Document.ColorSpace = ColorSpace
+Document.ColorSpaceMap = ColorSpaceMap
 
 // override getting the id to make sure it's fine if we have an MSDocument
 Document.define('id', {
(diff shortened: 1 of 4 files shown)


jedrichards:
Need to add docs and tests, but am I on the right track?

bohemian-coding-danger:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  - 📚 its docs
- 🧪 its tests

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 1d73769b99363c361de762197d964c8b77f907f9
</p>

mathieudutour:
That's the right track. When you have an enum, I usually have this pattern: https://github.com/BohemianCoding/SketchAPI/blob/develop/Source/dom/style/Fill.js#L118-L133 to return a human readable string instead of a number

jedrichards:
@mathieudutour Ready for another look. There is a difference between assigning and converting the color space, so it made sense to me to add a dedicated method `changeColorSpace` that allowed the user to choose between the two behaviours, and avoid using a setter and making the `colorSpace` prop read only. See what you think ...

Source/dom/models/Document.js
```diff
@@ -389,6 +421,8 @@ if (typeof MSDocument !== 'undefined') {
 }
 
 Document.SaveMode = SaveModeType
+Document.ColorSpace = ColorSpace
+Document.ColorSpaceMap = ColorSpaceMap
```
mathieudutour: no need to expose the map

jedrichards:
@mathieudutour Yep, I down with it! :smile:

docs/api/Document.md
```diff
@@ -221,6 +222,29 @@ document.close()
...
+
+// Pass true as an optional second argument
+// to convert instead of assign
+document.changeColorSpace(ColorSpace.P3, true)
+log(document.colorSpace === ColorSpace.P3) // true
+
+// Alternatively, use the property setter (the behaviour
+// here is to always assign the color space)
+document.colorSpace = ColorSpace.P3
+log(document.colorSpace === ColorSpace.P3) // true
```
mathieudutour: let's use `console.log`

Source/dom/models/Document.js
```diff
@@ -374,6 +386,26 @@ export class Document extends WrappedObject {
 
     msdocument.close()
   }
+
+  changeColorSpace(colorSpace, convert = false) {
```
mathieudutour: I'm wondering if it should be `convertColorSpace` and `convert` should just be `true` :thinking:

Is there a case where you'd use this instead of the `colorSpace` setter?

docs/api/Document.md
jedrichards: 👍

Source/dom/models/Document.js
jedrichards: I was thinking it could be awkward to have two different apis for assigning vs. converting. By having assigning and converting available via the same method you can switch the behaviour with a one liner:

```
doc.changeColorSpace(ColorSpace.P3, shouldConvert)
```

Rather than

```
if (shouldConvert) {
  doc.convertColorSpace(ColorSpace.P3)
} else {
  doc.colorSpace = ColorSpace.P3
}
```

🤔

mathieudutour: is there a use-case where `shouldConvert` would be dynamic? Tbh I'm still a little be confused why you'd not want to always convert?

jedrichards: I chatted to the Mac devs a bit about this. It sounds like converting is a potentially destructive act when switching between managed color profiles (it walks the document and mutates all the color values), but assigning just sets the property. So in a sense the sort of switch you want to do, assigning vs converting, may depend on the current color profile value ... hence my thinking of adding the flag to the method call.

Basically reproducing some of the similar flow available by the UI:

image

Beyond that, from the point of view of a designer using Sketch, the pros and cons of assigning vs converting are currently beyond my understanding ...

mathieudutour: Ah, maybe ¯\_(ツ)_/¯. Let's keep it that way then

jedrichards: I suppose to be super explicit we could have an api like,

```
document.colorProfile // getter
document.colorProfile = 'P3' /// Assigning setter
document.assignColorProfile('P3')
document.convertToColorProfile('P3')
```

🤔

PR #537, opened 2019-06-10
Groups ref file

jedrichards:
Refs #525


reference-files/plugin.sketchplugin/Contents/Sketch/groups.js
```diff
@@ -0,0 +1,62 @@
...
+    parent: page,
+    frame: new Rectangle(0, 0, 200, 200),
+  })
+
+  // Nest a group within the artboard containing two additional layers
+  var group = new Group({
+    name: 'MyGroup',
+    parent: artboard,
+    layers: [
+      new Text({
```
mathieudutour: not that it's an issue but you don't have to create the object here, you could specify the json representation directly:
```
var group = new Group({
     name: 'MyGroup',
     parent: artboard,
     layers: [
      {
        type: 'Text',
        text: 'Hello World',
         style: {
           borders: [],
           textColor: '#f00',
           fontSize: 28,
           fontFamily: 'Arial',
           fontWeight: 10,
         },
       },
       {
         type: 'Shape',
         name: 'MyShape',
         frame: new Rectangle(0, 50, 50, 50),
         style: {
           opacity: 0.5,
           borders: [],
           fills: [{ color: '#0f0', fillType: Style.FillType.Color }],
         },
       }
})
```

You could even specify all of that in the Artboard constructor

jedrichards: Ah interesting. Good to know!

mathieudutour: same for `rectangle` actually: you don't need the constructor, you can pass an json object directly

PR #545, opened 2019-06-13
Prototypes ref file

jedrichards:
Refs #525


reference-files/plugin.sketchplugin/Contents/Sketch/prototypes.js
```diff
@@ -0,0 +1,58 @@
...
+    parent: page,
+    frame: new Rectangle(200, 0, 200, 200),
+  })
+
+  // Add prototyping link from Artboard1 to Artboard2
+  const link1 = new Text({
+    text: 'Go to page 2',
+    parent: artboard1,
+    flow: {
+      target: artboard2,
```
mathieudutour: maybe you could use a specific animation? Can be added later

PR #573, opened 2019-07-18
Fix typo in Find.md

KevinGutowski:
Should be two arguments for the find method. Not just one string passed in.


KevinGutowski:
oo I’m a dummy and now get you can pass in a many parameters for the selector

Issue #513, opened 2019-05-23
Improve SketchAPI release process


nthState:
## Build output

Build output is an archive (*.zip or *.tar.gz) containing:

- Compiled/bundled SketchAPI JS file(s), shipped in `Sketch.app/Contents/Resources/SketchAPI*.js`
- Core modules (Node.js polyfills), shipped in `Sketch.app/Contents/Resources/core-modules/**`

As part of CI, on every commit the artefact is generated and published to a unique uri (service used here is an implementation detail, but S3 is a candidate).

### Example artefact uris

For a commit:

`https://{s3-bucket-uri}/{commit-hash}.zip`

For convenience the latest commit on a branch is available via a stable uri based on the branch name:

`https://{s3-bucket-uri}/{branch-name}.zip`

## Tests

Aside from linting, type-checking and other tests etc. integration tests should be run against Sketch itself during CI.

On developer workstations this is easy to achieve by downloading or building the desired Sketch version, while on CI we'll need the ability to use a specific version Sketch - especially when trying to coordinate updates to SketchAPI that also require changes to Sketch. A suggestion for how this could work on CI by using a branch naming convention:

- Look for a matching branch on the Sketch repo
    - If found, use the Sketch build from that branch for the integration tests
    - If not found, use the version from `develop`

## Integration with Sketch build process

The Sketch build process will be need to be updated roughly as follows:

- Remove SketchAPI as a submodule
- Add configuration that defines the SketchAPI build artefact uri to be included
- Download the artefact from the uri and extract to `Resources/`

### Updates to SketchAPI that don't affect Sketch

1. Open a PR in SketchAPI repo
2. Provide QA steps in PR description to validate work (maybe using `require('__pr')` or downloading the custom Sketch build from Jenkins)
3. Merge when QA and tests pass, and gather the resulting artefact url
4. Open PR on Sketch repo and update build config to point at a new SketchAPI asset url
5. Sketch is built on the PR against updated asset url

### Batching updates to Sketch repo

If working on many SketchAPI features in tandem it may be convenient to batch SketchAPI updates into a single PR on the Sketch repo to integrate

### Updates to SketchAPI that need to coordinate with Sketch updates

1. PR is opened on Sketch repo for necessary Sketch updates
2. PR is opened on SketchAPI repo, with mirrored branch naming, so the SketchAPI CI picks up the correct Sketch version
[...]

PR #575, opened 2019-07-19
Remove case of API being built as submodule

christianklotz:
Ref #513 

Build artefacts are always saved to the build folder in the API repo.

Now that Jenkins uses the same `npm run build` to produce the zipped API bundle, including the API and core modules, this case of building the API in the context of a submodule within the Sketch repo is no longer relevant.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -4,13 +4,7 @@ const babelLoader = require('@skpm/builder/lib/utils/babelLoader').default({})
 
 const PRODUCTION = process.env.NODE_ENV !== 'development'
 
-// heuristic to know if we are inside the Sketch repo
-const IS_BC_BUILD = /\/Modules\/SketchAPI$/.test(__dirname)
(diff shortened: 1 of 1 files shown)


jedrichards:
Looks good. Is this the only change to the SketchAPI repo required? Do we need to coordinate with everything being ready on the Sketch side?

christianklotz:
@jedrichards no further coordination required, since the Sketch side already uses the artefact generated by Jenkins

mathieudutour:
Shouldn’t this have been merged only when the develop branch of Sketch doesn’t use it as a submodule? 

Now if for some reason we need to wait for Sketch 58 to remove it as a submodule, it’s going to be painful

christianklotz:
Fair point, didn't think of it. Another reason to get this over the line asap. The Sketch side is almost ready so hopefully this scenario can be avoided.

PR #591, opened 2019-08-07
Gen ref files for 56.3

### LINKED ISSUE #583: Create reference files for Sketch 56
There's been Sketch 56, 56.1 and 56.2 with the point releases really being patch releases only so it may be sufficient to just provide a general update for 56.

jedrichards:
Closes #583 

- Adds ref files for 56.3
- Tweak scripts and dir structure so that the master ref files plugin is copied and used for each new version, in case fixes/changes are required in future


jedrichards:
I suppose the idea is that we're still able to point at the original plugin source that generated a given set of reference files, even if we create new versions of the plugin in future (e.g. add a new file type, or amend the plugin to improve the output, or whatever).

There are zero changes to the plugin or output in this PR if that helps things? The noise is just moving files to different folders, and duplicating the plugin for the older versions to keep their directory shapes the same as freshly generated files.

I can remove the new 56.3 files from this PR if you like and do them in a separate one?

mathieudutour:
Right but if you improve the output of the plugin (eg. change the document it's generating), it'd be best to regenerate it for all the versions otherwise the diff will be misleading.

If we do have some changes depending on sketch versions (because of a breaking change of some sort for example), then I'd say it's best to make same explicit and have different code paths in the same plugin. After all, the goal of all this is to make it easier to understand how the file is changing across versions

jedrichards:
Yep you might be right, I don't have a strong opinion either way really. Happy to revert to one master plugin, if you think that makes more sense.

The only thing I'd add is that we've been getting a lot of mileage from just the static list of reference file JSON, e.g. clicking on a JSON file to inspect the output, haven't felt like an explicit "diff feature" has been critical to the core usefulness so far.

Any opinions here @christianklotz ?

mathieudutour:
> we've been getting a lot of mileage from just the static list of reference file JSON, e.g. clicking on a JSON file to inspect the output, haven't felt like an explicit "diff feature" has been essential to the core usefulness so far.

That's true but maybe it's because it's not available yet?

christianklotz:
If there are no changes in the plugin for now we can revisit the version handling at a later stage and keep the single un-versioned plugins the same for now. We're still learning from going through release to release with this.

Key priority is providing reference files for the Sketch releases since this is how users usually discover the features and changes.

For context, before we started providing reference files we discussed both options, tagged released and organising content in version folders. IIRC tagged releases were discarded at the time because technically speaking diffs between versions would mostly be object/layer ids since they're autogenerated but aren't _actual_ diffs so it felt like we would eventually process these reference files anyway for a better.

As for changing plugin output, I understand the point but retrospectively providing reference files would:

1. Either make the case for organising output in version folders
1. Or require rewriting git history when using tagged releases

The reference files are not considered "feature complete" but we try to provide 

_**TLDR;** Just update reference files for now, keep plugins, let's add new issue for versioning_

mathieudutour:
> tagged releases were discarded at the time because technically speaking diffs between versions would mostly be object/layer ids since they're autogenerate

I'd say it's not a good reason to discard it. It's trivial to match UUID and replace them with a fixed string. I'd argue (and already did in a previous PR) we should do that anyway

scripts/generate-reference-files-md.js
```diff
@@ -28,7 +28,15 @@ const refFilesUrl = `${githubUrl}/tree/develop/reference-files`
 const versions = readdirSync(dir)
   .map(v => ({ original: v, coerced: semver.coerce(v) }))
   .filter(v => !!v.coerced)
-  .sort((a, b) => semver.lt(a.coerced.version, b.coerced.version))
+  .sort((a, b) => {
```
mathieudutour:
```suggestion
  .sort((a, b) => semver.compare(a.coerced.version, b.coerced.version))
```

jedrichards: Good catch.

PR #592, opened 2019-08-08
process is now an EventEmitter

mathieudutour:
It can now be used to trigger and catch `uncaughtException` for example


mathieudutour:
there wasn't any yet

christianklotz:
Ok, no worries. Next time, let's ensure to add an issue before creating a PR.

PR #615, opened 2019-08-29
Bump set-value

bomberstudios:
Fixes the vulnerability reported in https://github.com/BohemianCoding/SketchAPI/network/alert/package-lock.json/set-value/open


jedrichards:
I've checked the set-value repo and I don't think there're are any breaking api changes changing to version `2.0.1` so I reckon this is safe to merge. As a follow on task we should update dependencies to latest versions sometime too (created a note in the backlog).

Issue #316, opened 2018-12-20
Add more UI input types


KevinGutowski:
It currently looks like this:
<img width="532" alt="screen shot 2018-12-19 at 11 39 56 pm" src="https://user-images.githubusercontent.com/4199296/50270736-6ba7d280-03e7-11e9-8c5b-0cbfa595ce36.png">

It doesn't quite clearly communicate what value you are setting. Perhaps something like the image below would be a bit better as it would let the user at least find out what the min/max values are as well as being able to clearly see what value they have selected.

![slider inputcombo](https://user-images.githubusercontent.com/4199296/50270830-b6c1e580-03e7-11e9-8185-996e57071b3d.png)

mathieudutour:
Indeed, I didn't document it for that reason yet.

In addition to the extremum values, it'd be nice to show the steps (if set)

KevinGutowski:
Ahh gotcha, I won't include it in the post I'm writing up for SketchPlugins about the 53 Beta api changes. Thanks!

KevinGutowski:
Ah, yeah I did see that the other UI inputs commented as todos. Looking at the controls here ([https://developer.apple.com/documentation/appkit/views_and_controls?language=objc](https://developer.apple.com/documentation/appkit/views_and_controls?language=objc)) maybe a Date Picker or Segmented Control? Though, not sure if that is really too useful.

Not entirely sure about the core intention of these but it could be nice to be able to use multiple inputs within a modal (or possibly able to mix and match them). Perhaps thats something to consider at a later time.

> PS: if you want to link to the doc of the develop branch…
Ahh, great! Thanks. 👍

lpsBetty:
It would be awesome to have radio buttons when using `selection`. There is no setting yet to achieve this right?

KevinGutowski:
@lpsBetty Correct, but that should be straightforward to add - I'll see if I can get to adding it in this weekend.

christianklotz:
Closing as there are no plans to add this at this point. However we're looking into better UI integration for 3rd party development…

PR #622, opened 2019-09-17
Kevin/radio options

KevinGutowski:
[WIP]

From feedback in #316

By specifying the radio option variant for `INPUT_TYPE.selection` a developer can change the selection style.

Example:

Will need to pass extra parameter `radioStyle:true`

``` javascript
let UI = require('sketch').UI

UI.getInputFromUser("What's your favorite design tool?", {
  type: UI.INPUT_TYPE.selection,
  possibleValues: ['Something really long and aweful goes here and not sure what happens', 'Sketch', 'Hello'],
  radioStyle: true,
}, (err, value) => {
  if (err) {
    // most likely the user canceled the input
    console.error(err)
    return
  }
  console.log("selected value is: " + value)
})
```

Current concerns:

- [ ] Unsure what the best way to specify the UI type for selection (dropdown vs radio options). Want something that won’t break old code but allows for future expansion.
- [ ] Last? radio option can be highlighted
Sep-17-2019 01-32-13
- [ ] Had to disable linter warning on line [278](https://github.com/BohemianCoding/SketchAPI/compare/develop...KevinGutowski:kevin/radio_options?expand=1#diff-1165cabb99c47300408a4c1126da287fR278) Maybe try not using a for loop?
- [ ] Truncate clipped text

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -174,6 +174,7 @@ globals:
   NSStackView: false
   NSUserInterfaceLayoutOrientationVertical: false
   NSLayoutAttributeLeading: false
+  NSStackViewGravityLeading: false
 
 rules:
   ###########
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -168,6 +168,12 @@ globals:
   NSTextView: false
   NSMakeSize: false
   NSPopUpButton: false
+  NSButton: false
+  NSRadioButton: false
+  NSOnState: false
+  NSStackView: false
+  NSUserInterfaceLayoutOrientationVertical: false
+  NSLayoutAttributeLeading: false
 
 rules:
   ###########
--- a/Source/ui/UI.js
+++ b/Source/ui/UI.js
@@ -172,7 +172,6 @@ export function getInputFromUser(messageText, options, callback) {
       break
     }
     case INPUT_TYPE.selection: {
-
       console.log('triggered')
 
       if (!util.isArray(options.possibleValues)) {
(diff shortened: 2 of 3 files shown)


bomberstudios:
This looks awesome, thanks a lot for taking the time to build it!

We'll take a look at this when we have a minute (we're a bit busy with the 58 release at the moment 😅 )

KevinGutowski:
No worries! I’ll be cleaning it up a bit more over the next few days. I realized that I should probably extend this to also support a multiselect with checkboxes which slightly changes how I imagine interacting with the API should look like.

jedrichards:
@KevinGutowski I only found out about them myself recently, but the new(ish) [draft PRs](https://github.blog/2019-02-14-introducing-draft-pull-requests/) seem like a nice way to handle a PR in progress like this, avoiding the manual "WIP" labels and so on.

KevinGutowski:
Ahh thanks for the tip! I'll try that out. Mostly wanted to get this PR out so that others might be able to point me in the right direction.

jedrichards:
@KevinGutowski Done some spring cleaning in the repo and updated some dependencies. You might want to update vs. `develop` - not sure where your package-lock conflict is coming from.

The `no-plusplus` eslint rule seems a bit silly to me, feel free to disable it in the main eslintrc file. Otherwise you could just rewrite your loop definition without using `++`?

```
for (let i = 0; i < accessory.subviews().length; i=i+1) {
```

KevinGutowski:
Gonna close this Pull request since I haven’t worked on this in a while. Might pick it back up another time though.

PR #664, opened 2019-11-07
Always rebuild site on Netlify

bomberstudios:


netlify.toml
```diff
@@ -2,4 +2,3 @@
 	base    = ""
 	publish = "docs-website/_site"
 	command = "cd docs-website && bundle install && jekyll build"
-	ignore = "git diff --quiet HEAD^..HEAD -- docs docs-website"
```
jedrichards: Aha! 😅

bomberstudios: I think I finally have it!

PR #702, opened 2020-02-03
Fix sketchtool location

christianklotz:
The old location still has a symlink but actual binary is in `Sketch.app/Contents/MacOS/sketchtool`


jedrichards:
Aha interesting, I should update the code that generates the reference files with sketchtool too I guess https://github.com/sketch-hq/sketch-reference-files/blob/master/scripts/generate-ref-files.ts#L34

PR #704, opened 2020-02-11
Release 63 to master

jedrichards:


christianklotz:
Should this go into `master` first and then merge `master` into `develop`?

jedrichards:
@christianklotz I'm not sure, what would be the benefit?

christianklotz:
@jedrichards having the public release state on master. If there are no differences between `develop` and `master` then there won't be a problem but we certainly want the website to include latest release 🤔

jedrichards:
@christianklotz I see what you mean, I switched the base branch.

PR #707, opened 2020-02-19
Update Jekyll to v4.0.0

bomberstudios:
(WIP, just pushing to check if everything works as expected on Netlify)


bomberstudios:
> why did the deployment get cancelled.

No idea. I'll make an empty commit to force a redeploy

> Can we rejig to master instead of develop?

Sure thing

bomberstudios:
New hypothesis: maybe the deploy is cancelled by Netlify because it sees no new content on the site itself?

Will edit a page and see if that's it…

PR #723, opened 2020-04-13
Added note about removing override values

bomberstudios:


docs/api/SymbolInstance.md
```diff
@@ -78,7 +78,7 @@ A new [Group](#group) or `null`
 instance.setOverrideValue(instance.overrides[0], 'overridden')
 ```
 
-Change the value of the override.
+Change the value of the override. To remove an override (set a Text override to its default value, set a Symbol Instance override to `No Symbol`, etc…), set its value to an empty string (`''`).
```
christianklotz, line 81: Would it make more sense to move this information into the `value` parameter section since it's already mentioning details on the types, etc. It can be improved, perhaps with a list for text, image and symbols? Also helps clarifying if `''` works for images or not, since it states `NSImage – required` 🤷

bomberstudios: Turns out setting a bitmap override to `''` crashes Sketch, so clearly this needs a bit more work 😅

PR #744, opened 2020-05-12
Spring clean

### LINKED ISSUE #743: Rationalize tooling and scripts

jedrichards:
Resolves #743


Source/dom/models/__tests__/ImportableObject.test.js
```diff
@@ -1,51 +1,49 @@
 /* globals expect, test */
-import { isRunningOnJenkins, outputPath } from '../../../test-utils'
+import { outputPath } from '../../../test-utils'
 import { Library, Document, Artboard, Text, SymbolMaster } from '../..'
 
 // some tests cannot really run on jenkins because it doesn't have access to MSDocument
```
christianklotz, line 5: This comment can go 🎉

package.json
```diff
@@ -1,76 +1,31 @@
 {
   "name": "sketch-api",
   "private": true,
-  "version": "2.0.0",
+  "version": "0.1.0",
```
christianklotz, line 4: I know it doesn't really matter since it doesn't get published to npm. Is it even necessary/required to be included?

package.json
```diff
@@ -1,76 +1,31 @@
...
   "author": "Sam Deane",
```
christianklotz, line 5: It's about time to change this

```suggestion
  "author": "Sketch",
```

package.json
jedrichards: I think commands like `npm install` fail if `version` isn't set to _some_ valid semver. The default value for `version` in a fresh package.json created by npm is `1.0.0`. I'll set it to that, since I think it's the least surprising/misdirecting for a project that is technically unversioned.

mathieudutour: That's a change that could impact plugins: the API version is exposed in the API. It was 2.0.0 because there was an API 1.0.

package.json
```diff
@@ -25,7 +25,7 @@
     "prettier:base": "prettier --write",
     "prettify": "find . -name \"*.ts\" -o -name \"*.js\" | grep -v -f .gitignore | grep -v docs-website | xargs npm run prettier:base",
     "lint-staged": "lint-staged",
-    "postinstall": "node ./scripts/install-pre-commit.js && node ./scripts/inject-new-api-in-tester.js && rm -rf .awcache || true && cd core-modules && npm install",
```
mathieudutour, line 28: not sure what you mean by "not doing anything" but it is doing something: updating the test runner to use the API being worked on instead of the one shipping in Sketch (so you don't have to mess with your defaults)

jedrichards: Hmm - I checked the script, and it looked like the string it was trying to find and replace was no longer in the file it was operating on. So seemed like it wasn't having any effect ... did I get that wrong?

jedrichards: E.g. `Document = require('sketch')` isn't in https://github.com/skpm/skpm/blob/master/packages/test-runner/test-runner.sketchplugin/Contents/Sketch/tests-template.js

mathieudutour: Yeah should be `sketch = require('sketch')` instead of `Document = require('sketch')` (https://github.com/skpm/skpm/blob/master/packages/test-runner/test-runner.sketchplugin/Contents/Sketch/tests-template.js#L3).
Still useful IMO

jedrichards: After making the change I confirmed that I could still run the tests, make changes to the tests and source code and see updated results etc. It all seemed to work as expected.

mathieudutour: I'm guessing that's because you have set the `SketchAPILocation` which is fine for you, but dangerous for external contributors (because it won't reset and so they will always use an outdated version of the API)

jedrichards: It's confusing because even without that script the following workflow worked:

- Update source code
- Write a test against change
- Run `skpm-test` and see expected test results

I didn't have to mess with any `defaults write` stuff ...

... is that workflow impossible then in your opinion? If so I'll double check, its feasible I could have got into a muddle ...

mathieudutour: I mean it depends on what you are updating - but the tests will get a `Document` made with the API shipped with Sketch if you don't update your defaults (or have this script)

jedrichards: Hmm I tested with `Document`. I'll test again. But to be clear, that `postinstall` script hasn't been doing anything for over a year at least anyway, since it was searching for the wrong string.

I'll look into it ...

mathieudutour: sure, but nobody worked on it for a year 😄

package.json
jedrichards: Ah I see 👍

package.json
jedrichards: I can't reproduce any problem here. Here are my steps:

Confirm no default set,

```
$ defaults read com.bohemiancoding.sketch3 SketchAPILocation
2020-05-12 15:02:39.533 defaults[89939:4560432] 
The domain/default pair of (com.bohemiancoding.sketch3, SketchAPILocation) does not exist
```

Clean checkout

```
$ git clone git@github.com:sketch-hq/SketchAPI.git
$ cd SketchAPI
$ git checkout feature/spring-clean
$ npm install
```

Add this to `Document`

```
Document.define('foo', {
  exportable: true,
  importable: false,
  get() {
    return 'foo'
  },
})
```

Add this to `Document.test.js`

```
test('foo', (context, document) => {
  expect(document.foo).toBe('bar')
})
```

Then run

```
npm run test
```

And I get the expected test failure, showing latest code from `Document.js` and `Document.test.js` was picked up.

```
Document › foo

    expect(received).toBe(expected)
    
    Expected value to be (using Object.is):
      'bar'
    Received:
      'foo'
```

So struggling to see what could be improved here ...

mathieudutour: Ah that's because I found another way to do it (in the webpack config which is used by the test runner)

jedrichards: Ah great. I'm glad you did, it works pretty nice!

PR #774, opened 2020-06-09
Fix duplicate anchor

KevinGutowski:
There were two anchors for resizing a symbolInstance to match the auto layout (“both were `#trigger-a-smart-layout`. I tweaked the heading so that the anchors weren’t matching anymore.


christianklotz:
Well spotted :) One thing, can you make the change off `master` instead of `develop` so it can go up straight away?

KevinGutowski:
Done!

PR #798, opened 2020-09-06
Bump serialize-javascript

bomberstudios:
This is a manual PR.

As discussed in Slack, I've checked that:

- `npm run-script build` still works, and builds the packages correctly
- the output of the build command is not affected by the package update, and only `package.json`, `package.lock` and `node_modules` are changed (there are minor changes in the built tests, but that part of the build has never been deterministic so that's expected)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5068,9 +5068,12 @@
       "integrity": "sha512-b39TBaTSfV6yBrapU89p5fKekE2m/NwnDocOVruQFS1/veMgdzuPcnOM34M6CwxW8jH/lxEa5rBoDeUwu5HHTw=="
     },
     "serialize-javascript": {
-      "version": "2.1.2",
-      "resolved": "https://registry.npmjs.org/serialize-javascript/-/serialize-javascript-2.1.2.tgz",
-      "integrity": "sha512-rs9OggEUF0V4jUSecXazOYsLfu7OGK2qIn3c7IPBiffz32XniEp/TX9Xmc9LQfK2nQ2QKHvZ2oygKUGU0lG4jQ=="
+      "version": "3.1.0",
+      "resolved": "https://registry.npmjs.org/serialize-javascript/-/serialize-javascript-3.1.0.tgz",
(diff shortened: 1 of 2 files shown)


christianklotz (commented):
Rather than throwing `serialize-javascript` in here it seems like updating `test-runner` would be the better thing to do since the vast majority of warnings stem from that package 🤔

jedrichards:
@bomberstudios Yeah I tend to agree with @christianklotz on this one. I couldn't find the Slack convo mentioned above, but generally speaking (and this counts for the other recent security related PRs similar to this one) if at all avoidable, we ideally shouldn't fix security problems by introducing new top level dependencies that aren't used in actual app code.

In this case `serialize-javascript` is a transitive dependency meaning it just gets installed because its a dependency of a dependency. In future that could change, and we'd just be left dangling unused top level dependencies. Better to keep top level dependencies as an accurate picture of what's used by app code, and let the transitive dependencies get managed by npm or yarn, however they see fit.

Unfortunately that means each of these sorts of fixes usually has to start with some detective work to find out why transitive dependencies are in node_modules.

`npm ls` tells us `serialize-javascript` is in node_modules courtesy of `terser-webpack-plugin`:

```
$ npm ls serialize-javascript
sketch-api@2.0.0 /Users/jedrichards/Repos/SketchAPI
└─┬ webpack@4.39.3
  └─┬ terser-webpack-plugin@1.4.3
    └── serialize-javascript@2.1.2
```

Doing some [digging](https://github.com/webpack/webpack/issues/11309) tells us that the problem is fixed in Webpack 5, but that's still in pre-release, and the Webpack team aren't backporting the fix to the version of `terser-webpack-plugin` used in Webpack 4.

There's a decent fix mentioned as a [comment](https://github.com/webpack/webpack/issues/11309#issuecomment-674128535) though,

- Delete the entry for `terser-webpack-plugin` from `package-lock.json`
- Re-run `npm install`

That'll allow npm to naturally figure out a more modern version of the `serialize-javascript` transitive dependency that satisfies the requirements of the dependency graph, and boom the high severity security warnings are gone 🎉

bomberstudios:
Going to close this PR and address this properly on a new one

PR #799, opened 2020-09-16
Feature/32044 color variables api for plugins

bomberstudios:
For https://github.com/sketch-hq/Sketch/issues/32044

### DIFF
--- a/Source/dom/models/Document.js
+++ b/Source/dom/models/Document.js
@@ -698,10 +698,8 @@ Document.define('swatches', {
     if (!this._object) {
       return []
     }
-    const documentData = this._getMSDocumentData()
-    return toArray(documentData.sharedSwatches().swatches()).map((a) =>
-      Swatch.fromNative(a)
-    )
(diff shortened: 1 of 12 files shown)


Source/dom/models/Document.js
```diff
@@ -686,6 +687,59 @@ Document.define('gradients', {
...
+    }
+  },
+  removeItem(index){
+    if (this.isImmutable()) {
+      return
+    }
+    const assets = this._getMSDocumentData().sharedSwatches()
+    const swatch = assets.swatchAt(index)
+    assets.removeSharedObjects([swatch])
+  },
```
christianklotz: If swatches are internally an unordered sequence it may be better to implement it as a set instead of an array 🤔

bomberstudios: AFAIK they are ordered (or so I wrote on my dev notes: "Internally, swatches are returned in order, and apparently the order does not change when changing a Swatch name")

christianklotz (approved):

PR #801, opened 2020-09-23
Adds some randomness to output paths for test assets

bomberstudios:
Connect to https://github.com/sketch-hq/Sketch/issues/32823

### DIFF
--- a/Source/test-utils.js
+++ b/Source/test-utils.js
@@ -56,7 +56,8 @@ export function createSharedStyle(document, Primitive, style) {
 }
 
 export function outputPath() {
-  const output = `${NSTemporaryDirectory()}SketchAPI/`
+  const uuid = NSUUID.UUID().UUIDString()
+  const output = NSTemporaryDirectory().stringByAppendingPathComponent('SketchAPI/').stringByAppendingPathComponent(uuid).stringByAppendingPathComponent('/')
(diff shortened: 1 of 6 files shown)


Source/test-utils.js
```diff
@@ -56,7 +56,8 @@ export function createSharedStyle(document, Primitive, style) {
 }
 
 export function outputPath() {
-  const output = `${NSTemporaryDirectory()}SketchAPI/`
+  const uuid = NSUUID.UUID().UUIDString()
+  const output = NSTemporaryDirectory().stringByAppendingPathComponent('SketchAPI/').stringByAppendingPathComponent(uuid).stringByAppendingPathComponent('/')
   fs.mkdirSync(output, { recursive: true })
   return output
```
christianklotz: Given that we're using `outputPath/…` everywhere I think we can drop the last appended `/`. Then double check we _are_ actually using `/` in the places where we use the path since the topic has come up a few times recently. Maybe we can improve readability a little with some line breaks, too?

```suggestion
  const uuid = NSUUID.UUID().UUIDString()
  const path = NSTemporaryDirectory()
    .stringByAppendingPathComponent('SketchAPI/')
    .stringByAppendingPathComponent(uuid)
  
  fs.mkdirSync(path, { recursive: true })
  return path
```

christianklotz (approved):

PR #808, opened 2020-10-13
Improve testing

christianklotz:
Resolve https://github.com/sketch-hq/Sketch/issues/33055

PR #828, opened 2021-02-26
Feature/35809 cherrypick api testing

christianklotz:
Same changes from #808 but for Sketch 71.


christianklotz:
Closing as this will go into a later version

PR #844, opened 2021-08-27
Define property Flow.maintainScrollPosition

royalrex:
Add the Flow.maintainScrollPosition property.

Sketch prototyping allows setting this property through the UI, but it's not available through the API.

I have added it to the best of my ability, and loading the API locally I can use it as expected in my plugin.

I have added the new property to Flow tests, as well as a specific test.

### DIFF
--- a/Source/dom/models/Flow.js
+++ b/Source/dom/models/Flow.js
@@ -113,5 +113,17 @@ Flow.define('animationType', {
   },
 })
 
+Flow.define('maintainScrollPosition', {
+  get() {
+    return Boolean(this._object.maintainScrollPosition());
+  },
+  set(maintainScrollPosition) {
+    if (this.isImmutable()) {
+      return
+    }
+    this._object.setMaintainScrollPosition(maintainScrollPosition)
+  },
+})
+
 // override the WrappedObject id
 delete Flow[DefinedPropertiesKey].id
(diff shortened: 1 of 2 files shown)


royalrex:
@InvisiblePixels hope everything is good here. Let me know if I need to add something to docs (https://developer.sketch.com/reference/api/#flow)

bomberstudios:
Thank you very much for the PR!

Code is pretty much ready to merge, so I’ll just add a couple of comments:

- I’d maybe add a test to make sure that changing the value of the property from `false` to `true` does actually work. I know that might sound a bit offensive, because the actual code we’re testing couldn’t be simpler, but since we’re dealing with mutable _and_ immutable properties, it never hurts to double check.
- I would also try to add some kind of check to make sure we’re not setting the property to a non-boolean value. Sketch’s codebase may be making some assumptions about the value, and again it doesn’t hurt to make sure we’re not passing something unexpected.

Meanwhile, I will enable the automatic checks for this PR, which GitHub has decided to disable for new contributors, so we can all be confident that things work as expected.

Again, thanks a lot for this!

royalrex:
Thank you for your review.

I will implement your suggestions when I get some extra time (expect this week or next) - and sorry for the long response time.

royalrex:
Hi @bomberstudios 

Got swamped with other work, sorry for the long delay. I have added the suggested test as well as a type check.
Can see someone else is also in need of this pull request (romannurik/Sketch-FlowExporter#51). 

Let me know if you have any other suggestions to improve the quality of this PR.

Best.

bomberstudios (approved):

PR #846, opened 2021-09-03
Code Review: #38875 Added missing case for Text tests and changed Group tests to look for truthiness rather than 1

acf:
See also https://github.com/sketch-hq/Sketch/issues/38875

### DIFF
--- a/Source/dom/assets/__tests__/GradientAsset.test.js
+++ b/Source/dom/assets/__tests__/GradientAsset.test.js
@@ -20,15 +20,15 @@ test('should create gradient asset from MSGradientAsset', (_context, document) =
 })
 
 test('should create gradient asset with name', (_context, document) => {
-  const gradient = {
-    name: 'My Gradient',
-    gradient: {},
-  }
-
-  document.gradients.push(gradient)
-  expect(document.gradients.length).toEqual(1)
-
-  const asset = document.gradients[0]
-  expect(asset.gradient.type).toBe('Gradient')
-  expect(asset.name).toBe('My Gradient')
+//   const gradient = {
+//     name: 'My Gradient',
+//     gradient: {},
+//   }
+// 
+//   document.gradients.push(gradient)
+//   expect(document.gradients.length).toEqual(1)
+// 
+//   const asset = document.gradients[0]
+//   expect(asset.gradient.type).toBe('Gradient')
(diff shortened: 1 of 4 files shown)


Source/dom/assets/__tests__/GradientAsset.test.js
```diff
@@ -20,15 +20,15 @@ test('should create gradient asset from MSGradientAsset', (_context, document) =
...
+//     name: 'My Gradient',
+//     gradient: {},
+//   }
+// 
+//   document.gradients.push(gradient)
+//   expect(document.gradients.length).toEqual(1)
+// 
+//   const asset = document.gradients[0]
+//   expect(asset.gradient.type).toBe('Gradient')
+//   expect(asset.name).toBe('My Gradient')
```
christianklotz: Are these changes unintentional? I am assuming this test needs fixing but will be addressed later? If that's the case, I'd suggest to leave it as is instead of commenting out.

Source/dom/layers/__tests__/Text.test.js
```diff
@@ -67,13 +67,14 @@ test('should change the line spacing behavior', () => {
   })
 
   // default to constant baseline
-  expect(text.lineSpacing).toBe(Text.LineSpacing.natural)
+  expect(text.lineSpacing).toBe(Text.LineSpacing.system)
```
christianklotz: The `// default to constant baseline` comment needs updating, or perhaps better removal. Seems like it's not adding much (besides being factually wrong)

Source/dom/assets/__tests__/GradientAsset.test.js
acf: Oh damn yes!  Sorry!

Source/dom/layers/__tests__/Text.test.js
acf: 👍

Source/dom/assets/__tests__/GradientAsset.test.js
acf: The perils of editing in bbedit and committing on the command line

Source/dom/layers/__tests__/Text.test.js
acf: updated the comment to hopefully explain better

christianklotz (approved):

PR #854, opened 2021-09-28
Code Review: Do not pre-select edit point on entering vector edit mode (#32796)

acf:
Connect sketch-hq/Sketch#32796

### DIFF
--- a/Source/dom/models/__tests__/CurvePoint.test.js
+++ b/Source/dom/models/__tests__/CurvePoint.test.js
@@ -53,7 +53,8 @@ test('should be able to tell if a point is selected)', () => {
   expect(shape.points[0].isSelected()).toBe(false)
 
   //select the first point  
-  shapeEventHandler.pathController.selectNext()
+  document.sketchObject
+    .eventHandlerManager()currentHandler.pathController.selectNext()
 
   //now selected true  
   expect(shape.points[0].isSelected()).toBe(true)
--- a/Source/dom/models/__tests__/CurvePoint.test.js
+++ b/Source/dom/models/__tests__/CurvePoint.test.js
@@ -33,11 +33,13 @@ test('should be able to modify a CurvePoint', () => {
   expect(p.point.toJSON()).toEqual({ x: 0.3, y: 0.4 })
 })
 
-test('switching to path edit mode does not select a point (#32796)', () => {
+test('should be able to tell if a point is selected)', () => {
   const document = new Document()
   const shape = new ShapePath({
     parent: document.pages[0],
   })
+  
+  // no point selected
   expect(shape.points[0].isSelected()).toBe(false)
 
   shape.selected = true
@@ -47,6 +49,14 @@ test('switching to path edit mode does not select a point (#32796)', () => {
     .eventHandlerManager()
     .switchToEventHandlerClass(MSShapeEventHandler.class())
 
+  //still no selection
   expect(shape.points[0].isSelected()).toBe(false)
+
+  //select the first point  
+  shapeEventHandler.pathController.selectNext()
+
+  //now selected true  
+  expect(shape.points[0].isSelected()).toBe(true)
+  
   document.close()
 })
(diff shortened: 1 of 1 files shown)


christianklotz (approved):

Source/dom/models/__tests__/CurvePoint.test.js
```diff
@@ -33,7 +33,7 @@ test('should be able to modify a CurvePoint', () => {
   expect(p.point.toJSON()).toEqual({ x: 0.3, y: 0.4 })
 })
 
-test('should show if a point is selected', () => {
+test('switching to path edit mode does not select a point (#32796)', () => {
```
jpedroso: @acf I don't think this change is right. 

The way I see it, this test was describing the expectation that the `CurvePoint` model reflects the `selected` state correctly when a certain set of actions is performed on the document. Correct me if I'm wrong. 

Now the test does a completely different thing. Your changes turn the test into a verification of application behavior when entering Shape Edit mode, which to me does not belong in SketchAPI tests.

In a hypothetical scenario, let's say that in the future the backing store for the `isSelected` property changes and it stops being computed correctly. This test will still happily pass.

The problem was, until now, the test assumed a point is immediately selected when switching handler to `ShapeEventHandler` and this changed. 

In my opinion, the solution for this test is to change the set of actions that lead to having a point selected after switching to "path edit mode". Specifically, we should be able to access the `shapeEventHandler`, and call `selectNext()` on its `pathController`. Something like the following pseudocode assuming the CocoaScript API allows.

```
shapeEventHandler.pathController.selectNext()
```

After this, the expectation can stay `expect(shape.points[0].isSelected()).toBe(true)`.

jpedroso (changes requested):

Source/dom/models/__tests__/CurvePoint.test.js
acf: fair point.  :-)

Source/dom/models/__tests__/CurvePoint.test.js
```diff
@@ -47,6 +47,6 @@ test('should show if a point is selected', () => {
     .eventHandlerManager()
     .switchToEventHandlerClass(MSShapeEventHandler.class())
 
-  expect(shape.points[0].isSelected()).toBe(true)
+  expect(shape.points[0].isSelected()).toBe(false)
```
jpedroso, line 53: After a chat with @christianklotz 🙏, finally found the motivation to fire VS Code and write the following instead of writing dumb js.

```suggestion
  // ask handler to select any point; by default, nothing's selected.
  document.sketchObject
    .eventHandlerManager()
    .currentHandler()
    .pathController()
    .selectNext(null)

  expect(shape.points[0].isSelected()).toBe(true)
```

acf: third time is the charm.  this is what I get for trying to do this in a hurry before a meeting.

Source/dom/models/__tests__/CurvePoint.test.js
```diff
@@ -47,6 +49,17 @@ test('should show if a point is selected', () => {
     .eventHandlerManager()
     .switchToEventHandlerClass(MSShapeEventHandler.class())
 
+  //still no selection
+  expect(shape.points[0].isSelected()).toBe(false)
+
+  //select the first point  
+  .eventHandlerManager()
```
jpedroso, line 56: Syntax is totally foreign to me but I take it that calling `.eventHandlerManager()` like this is equivalent to `document.sketchObject.eventHandlerManager()`, right?

Given the JavaScript proficiency levels among us, Mac devs, and for the sake of consistency, principle of least surprise, etc., maybe we should stick to the same syntax as the call above, or extract a local var. Wdyt?

Source/dom/models/__tests__/CurvePoint.test.js
```diff
@@ -47,6 +49,17 @@ test('should show if a point is selected', () => {
     .eventHandlerManager()
     .switchToEventHandlerClass(MSShapeEventHandler.class())
 
+  //still no selection
+  expect(shape.points[0].isSelected()).toBe(false)
```
jpedroso, line 53: This seems unnecessary because, again, it's assuming application behavior but I'm OK keeping it.

jpedroso (approved):
Thanks for changing the test and keeping the original intent. I assume the tests are now passing so I'm approving, but also left a couple of optional suggestions to improve test readability.

christianklotz (approved):

PR #855, opened 2021-09-28
Code Review: Allow proper update of a rectangle's corner radii from the API (#39183)

lrossi:
Connect sketch-hq/sketch#39183

Created by `bo pr`.

### DIFF
--- a/Source/dom/models/__tests__/CurvePoint.test.js
+++ b/Source/dom/models/__tests__/CurvePoint.test.js
@@ -87,8 +87,9 @@ test('should be able to tell if a point is selected)', () => {
   //still no selection
   expect(shape.points[0].isSelected()).toBe(false)
 
-  //select the first point  
-  .eventHandlerManager()
+  //select the first point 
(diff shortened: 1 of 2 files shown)


mikeabdullah (approved):

Source/dom/models/__tests__/CurvePoint.test.js
```diff
@@ -52,8 +87,9 @@ test('should be able to tell if a point is selected)', () => {
   //still no selection
   expect(shape.points[0].isSelected()).toBe(false)
 
-  //select the first point  
-  .eventHandlerManager()
```
lrossi, line 56: @acf API tests were failing after I merged develop and I _think_ there was a problem with this line. I fixed it in a way that I'm hoping matches the intention, could you please double-check if it makes sense? Thanks

acf: It does 🙈

acf (approved):

PR #861, opened 2022-01-25
Code Review: Fix SketchAPI tests that result from MSLayerArray removal (#40942)

pieteromvlee:
Connect sketch-hq/sketch#40942

Created by `bo pr`.

### DIFF
--- a/Source/dom/models/Selection.js
+++ b/Source/dom/models/Selection.js
@@ -22,7 +22,7 @@ export class Selection {
     Object.defineProperty(this, 'layers', {
       enumerable: true,
       get() {
-        const layers = toArray(this._object.selectedLayers().layers()).map(
+        const layers = toArray(this._object.selectedLayers()).map(
           wrapNativeObject
         )
         return layers
@@ -37,7 +37,7 @@ export class Selection {
     Object.defineProperty(this, 'length', {
       enumerable: true,
       get() {
-        return this._object.selectedLayers().layers().count()
+        return this._object.selectedLayers().count()
       },
     })
 


Source/dom/models/Selection.js
```diff
@@ -37,7 +37,7 @@ export class Selection {
     Object.defineProperty(this, 'length', {
       enumerable: true,
       get() {
-        return this._object.selectedLayers().layers().count()
+        return this._object.selectedLayers().count()
```
mikeabdullah, line 40: I'm trying to wrap my head around these. Are we:

- A) Changing the **API** to not use `layers()` and plug-ins using this will need to update too?
- B) Changing the implementation of something that we then _expose_ as API, so therefore plug-ins themselves don't care?

pieteromvlee: As I understand it, it's B cc @bomberstudios

mikeabdullah (approved):
Tentatively approving unless @bomberstudios says we've got this wrong. @lrossi also want to have a look to see if any setter methods here have a problem.

Source/dom/models/Selection.js
bomberstudios: I can confirm it’s B. You can tell because that line uses `_object`, which basically means “drop down to internal objects, this should not be used publicly”

bomberstudios (approved):

Source/dom/models/Selection.js
lrossi: Apologies for the delay. I had a deeper look at this, and it looks good to my eyes too. 👌

As long `selectedLayers()` is now returning the same elements that `selectedLayers().layers()` was, which seems to be the case, it should be transparent to callers, as others said.

Perhaps there is now no need for the extra `layers.layers` check in the corresponding setter (I'm not sure why it was there), but it doesn't look harmful, and it's probably safer to keep it.

lrossi (approved):

PR #862, opened 2022-03-03
Code Review: Fix SketchAPI target dependency in the Xcode project (#41714)

lrossi:
Connect sketch-hq/sketch#41714

Created by `bo pr`.


jnozzi (approved):
Until it's easier to specify inter-project dependencies within a workspace's targets and those targets' build phase scripts, this seems like the easiest option that doesn't sacrifice implicit dependencies and/or parallel builds.

PR #866, opened 2022-04-29
Get platform independent text alignment value

### LINKED ISSUE #860: TextAlignment mistake in sketch 80 on apple silicon
In sketch JSAPI and sketch-file-format-ts, text center alignment is value 2 and right is 1. 
But on apple silicon, When I set the text alignment to center, JSAPI returns 'right' and Objective-C api returns 1 too.
I unzipped the .sketch file, the alignment of the corresponding text is indeed 2.
When I run sketch by Rosetta, It works correctly.

klaaspieter:
NSTextAlignment uses different numerical values for different
architectures. Using the raw value means we incorrectly report text
alignment on M1 machines.

This fixes the issue by using `MSTextAlignmentConverter` used internally
by Sketch.

connects https://github.com/sketch-hq/Sketch/issues/39019

Closes https://github.com/sketch-hq/SketchAPI/issues/860

### DIFF
--- a/Source/dom/style/Text.js
+++ b/Source/dom/style/Text.js
@@ -188,7 +188,12 @@ export function defineTextStyleProperties(Style) {
       }
 
       updateParagraphStyle(this._object, (paragraphStyle) => {
-        const translated = TextAlignmentMap[mode]
+        // Get a platform independent raw text alignment value
+        // NSTextAlignment is represented differently on Intel and M1 chips
+        const translated =
+          TextAlignmentMap[
+            MSTextAlignmentConverter.archiveNSTextAlignment(mode)
+          ]
         paragraphStyle.setAlignment(
           typeof translated !== 'undefined' ? translated : mode
         )
--- a/Source/dom/style/Text.js
+++ b/Source/dom/style/Text.js
@@ -175,7 +175,11 @@ export function defineTextStyleProperties(Style) {
         return undefined
       }
 
-      const raw = paragraphStyle.alignment()
+      // Get a platform independent raw text alignment value
+      // NSTextAlignment is represented differently on Intel and M1 chips
+      let raw = MSTextAlignmentConverter.archiveNSTextAlignment(
+        paragraphStyle.alignment()
+      )
       return TextAlignmentReverseMap[raw] || raw
     },
     set(mode) {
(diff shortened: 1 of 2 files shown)


Source/dom/style/Text.js
```diff
@@ -175,7 +175,10 @@ export function defineTextStyleProperties(Style) {
         return undefined
       }
 
-      const raw = paragraphStyle.alignment()
+      // eslint-disable-next-line
```
lrossi: Another option is to add `MSTextAlignmentConverter` to the `.eslintrc`, which seems to be how Sketch's types are handled in the project.

Source/dom/style/Text.js
```diff
@@ -175,7 +175,10 @@ export function defineTextStyleProperties(Style) {
...
+      let raw = MSTextAlignmentConverter.archiveNSTextAlignment(
```
lrossi, line 180: May be worth adding a comment explaining why this is necessary, given the method name doesn't particularly help 🙂

lrossi:
@klaaspieter great job, only a couple of minor comments 👍

lrossi:
@klaaspieter all good and tested on M1, though it occurred to me that we need to do the reverse in the setter too 🙃

klaaspieter:
@lrossi added it to the setter as well. Can you take another look and make sure it works on M1?

Source/dom/style/Text.js
```diff
@@ -184,7 +188,12 @@ export function defineTextStyleProperties(Style) {
       }
 
       updateParagraphStyle(this._object, (paragraphStyle) => {
-        const translated = TextAlignmentMap[mode]
+        // Get a platform independent raw text alignment value
+        // NSTextAlignment is represented differently on Intel and M1 chips
+        const translated =
+          TextAlignmentMap[
+            MSTextAlignmentConverter.archiveNSTextAlignment(mode)
```
lrossi: @klaaspieter it doesn't seem to be working on M1: `mode` appears to be a string here (e.g. `"center"`), maybe we need to pass `TextAlignmentMap[mode]` to `archiveNSTextAlignment` instead? (The other way around)

klaaspieter: Should be fixed now 🤞🏻

klaaspieter:
When testing the setter please test both `center` and `NSTextAlignmentCenter` variants because it took quite some time to get both to work 😓

lrossi (approved):
👌

PR #867, opened 2022-04-29
Add watch script

klaaspieter:
Use Webpack watch mode to automatically watch all files within the
dependency graph for changes. If a file is updated webpack will
automatically recompile.

See https://webpack.js.org/guides/development/#using-watch-mode

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   "homepage": "https://github.com/sketch-hq/SketchAPI",
   "scripts": {
     "build": "./build.sh",
+    "watch": "webpack --watch",
     "test:build": "webpack --config webpack.tests.config.js --env.identifier=$npm_config_identifier --env.spec=$npm_config_spec",
     "lint": "eslint \"Source/**\"",
     "format-check": "prettier --check \"**/*.{js,json}\"",
--- a/README.md
+++ b/README.md
@@ -69,14 +69,15 @@ In addition to the API, this project also defines core modules to be included in
 
 The following npm scripts are available for development of the API.
 
-| Script                        | Description                             |
-| ----------------------------- | --------------------------------------- |
-| `npm run build`               | Build SketchAPI into the `build` folder |
-| `npm run test:build`          | Build integration test plugin           |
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -8,6 +8,7 @@
   "homepage": "https://github.com/sketch-hq/SketchAPI",
   "scripts": {
     "build": "./build.sh",
+    "watch": "webpack --watch",
```
lrossi, line 11: does this call the `build` command when recompiling (and thus the `build.sh` script) or is it equivalent to running `webpack` directly?

lrossi (commented):
Looks like a great addition, one question given my lack of JavaScript experience 🙂

package.json
klaaspieter: It is equivalent to calling `webpack` directly. I didn't see the line about `core_modules` in the `build.sh` 🙈 . Webpack automatically creates the build directory so we don't need the script for that.

Can you explain what this line does?

```
cp -R ./core-modules/node_modules/. ./build/core-modules
```

lrossi: > Can you explain what this line does?

@jedrichards might know? 🙏

jedrichards: Umm, this was from before my time at Sketch but it relates to the dependency between SketchAPI and the 3rd party [skpm](https://github.com/skpm) tool.

That `core-modules` folder includes a bunch of polyfills that attempt to make the JS environment inside Sketch plugins look and feel more like a traditional Node environment. For example, if you do `const fs = require('fs')` in a plugin, you get a module with the same API as Node's official `fs` module, but behind the scenes it's implemented with Cocoascript that works inside Sketch's plugin JS env.

You can see those 3rd party modules being mapped into the bundle as Webpack "externals" in the webpack config [here](https://github.com/sketch-hq/SketchAPI/blob/develop/webpack.config.js#L41).

klaaspieter: Okay. So just running the watch script will not work then? If so I'll close the PR 😞. Unless you know of an easy solution.

jedrichards: Nothing is easy with SketchAPI 😅

PR #871, opened 2022-05-27
Code Review: [SketchAPI] Support getting non-JSON values fro user defaults (#43719)

klaaspieter:
Fix `Setting.globalSettingForKey` returning `undefined` when the value in user defaults is not valid JSON. Also update some doc comments that incorrectly say that `globalSettingForKey` and `setGlobalSettingForKey` are the same as `defaults read` and `defaults write`.

Connect sketch-hq/sketch#43719


bomberstudios (approved):
Code looks good, and does what it says in the tin.

So, let’s get this merged

PR #892, opened 2022-07-08
PR: on feature/43561

opsGavin:
Connect sketch-hq/sketch#43561

Created by `bo pr`.

### DIFF
--- a/Source/dom/layers/SymbolInstance.js
+++ b/Source/dom/layers/SymbolInstance.js
@@ -75,6 +75,7 @@ export class SymbolInstance extends StyledLayer {
     if (this.isImmutable()) {
       return this
     }
+		this._object.ensureDetachHasUpdated()
     this._object.resizeToFitContentsIfNeededNoCache()
     return this
   }
--- a/Source/dom/models/Override.js
+++ b/Source/dom/models/Override.js
@@ -14,9 +14,10 @@ export class Override extends WrappedObject {
     if (!this.__symbolInstance) {
       return undefined
     }
+		this.__symbolInstance.sketchObject.ensureDetachHasUpdated()
     return toArray(
       this.__symbolInstance.sketchObject.overrideContainer().flattenedChildren()
-    ).find((x) => x.availableOverride() == this.sketchObject)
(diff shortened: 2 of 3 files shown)


Source/dom/models/Override.js
```diff
@@ -14,9 +14,10 @@ export class Override extends WrappedObject {
     if (!this.__symbolInstance) {
       return undefined
     }
+		this.__symbolInstance.sketchObject.ensureDetachHasUpdated()
     return toArray(
       this.__symbolInstance.sketchObject.overrideContainer().flattenedChildren()
-    ).find((x) => x.availableOverride() == this.sketchObject)
+    ).find((x) => x.availableOverride().overridePoint().name().isEqual(this.sketchObject.overridePoint().name()))
```
opsGavin, line 20: We don't have the same caching as we had before so the underlying objects could be different and the equality check could fail. Checking the override point name is safer

... but also we need to use `isEqual` rather than `==` 😖

Source/dom/layers/SymbolInstance.js
```diff
@@ -75,6 +75,7 @@ export class SymbolInstance extends StyledLayer {
     if (this.isImmutable()) {
       return this
     }
+		this._object.ensureDetachHasUpdated()
```
opsGavin, line 78: The optimal size is calculated by the detaching code, so this needs to complete before `resizeToFitContentsIfNeededNoCache` is called or it won't know the correct size.

opsGavin:
These changes have been introduced because we are enabling the symbol click-through flag `inHouse`. There are two changes of note.

The first is that under clickthrough we have changed how `availableOverrides` are calculated. Previously, this relied on the symbol instance being invalidated. We no longer do this and instead we now rely on the detached versions of the symbols. However, detaching is done on a background thread and may not have completed by the time the overrides are requested, so to get around that we need to call `ensureDetachHasUpdated()` which will block until the current detach completes. 

The second change with clickthrough is that, unlike before, when an override is selected the symbol instance isn't considered selected. Selecting overrides is now much more like selecting a normal layer. I've update the override selection tests to reflect this, and also updated the selection code to use the newer `Selection` and `SelectionItem` objects (`selectedOverrides` is deprecated and will go away soon).

acf (approved):
All seems sensible enough :)

